PR #69, opened 2013-03-23
support for 1 argument commands using PUT

kaytwo:
The unit tests passed, and my project using PUT for everything is still working, so I figured I'd send this over.


davidkhala:
@nicolasff 12 years ago, can we take it as stale?

PR #113, opened 2014-12-27
Fix handling of ports over 32767 by using 'int' instead of 'short'.

jplevyak:


nicolasff:
Sorry that it took so long to merge this. I'll also tag a new release later today. Thanks for the fix!

PR #133, opened 2016-06-15
Improve support for newer msgpack-c releases.

jamessan:
- Leverage msgpack's pkg-config information when available
- Enable building against newer (>= 1.2.0) msgpack-c releases,
  which had to make incompatible changes to accomodate changes
  in the msgpack spec itself.


jamessan:
Updated to also look for libmsgpackc in multi-arch library paths for msgpack-c releases which don't yet have pkg-config support (even though those are quite old).

jamessan:
Ping?

jollyroger:
@nicolasff do you plan to accept this PR? It  provides support for modern versions of the library and makes external patching unnecessary. FYI: Webdis exists in Debian archive with these patches for two years already.

PR #150, opened 2018-08-23
Pass tests with python-msgpack >= 0.3

### LINKED ISSUE #92: Failing tests on Ubuntu using python-msgpack v0.3
Nicolas, I've found that webdis fails msgpack-related tests using python-mspack with version 0.3.0. All tests pass with python-msgpack version 0.2.0 installed. However this package only is available in Debian and not Ubuntu. This affects recent Ubuntu versions(saucy and upcoming trusty), here is detailed output:

<pre>
python tests/basic.py
............F..FF.....
======================================================================
FAIL: test_error (__main__.TestMsgPack)
  File "tests/basic.py", line 145, in wrapper
    fn(self)
  File "tests/basic.py", line 192, in test_error
    self.assertTrue(isinstance(obj['UNKNOWN'], tuple))
AssertionError: False is not true

======================================================================
FAIL: test_list (__main__.TestMsgPack)
  File "tests/basic.py", line 145, in wrapper
    fn(self)
  File "tests/basic.py", line 183, in test_list
    self.assertTrue(obj == {'LRANGE': ('abc', 'def')})
AssertionError: False is not true

======================================================================
FAIL: test_set (__main__.TestMsgPack)
  File "tests/basic.py", line 145, in wrapper
    fn(self)
  File "tests/basic.py", line 157, in test_set
    self.assertTrue(obj == {'SET': (True, 'OK')})
AssertionError: False is not true

Ran 22 tests in 0.053s

FAILED (failures=3)</pre>

These errors do not appear for Debian jessie(current testing) and sid(unstable). Debian has two packages related to python and msgpack that are used in the test suite: `python-msgpack` with version 0.3.0 of the library and `msgpack-python` with version 0.2.0 of the same library. Ubuntu has just `msgpack-python` package that points to the 0.3.0 version. I assume version 0.2.0 will be dropped soon from Debian as well.

jollyroger:
Current test suite pass with python-msgpack v0.2 but fails with python-msgpack
v0.3 and later due to changes in the library. Modern distributions ship
python-msgpack > 0.3 for very long time ago. Closes #92

### DIFF
--- a/tests/basic.py
+++ b/tests/basic.py
@@ -154,7 +154,7 @@ class TestMsgPack(TestWebdis):
 		f = self.query('SET/hello/world.msg')
 		self.assertTrue(f.headers.getheader('Content-Type') == 'application/x-msgpack')
 		obj = msgpack.loads(f.read())
-		self.assertTrue(obj == {'SET': (True, 'OK')})
+		self.assertTrue(obj == {'SET': [True, 'OK']})
 
 	@need_msgpack
 	def test_get(self):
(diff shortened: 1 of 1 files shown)


nicolasff:
Integrated in 2496fb2515aadff6c5f270f224eb213557191aed. Thanks for the fix and sorry that it took so long, I should have merged this long ago.

PR #151, opened 2018-08-23
Print HTTP port number in the logs during startup.

jollyroger:
This provides a way to start webdis with dynamic port allocation and
discover HTTP port number by simply grepping logs i.e. without need to
use netstat or sockstat that are not available on some architectures.

Dynamic port allocation is a feature that can be used to run webdis
ad-hoc for testing purpose.

### DIFF
--- a/server.c
+++ b/server.c
@@ -25,6 +25,7 @@ socket_setup(struct server *s, const char *ip, int port) {
 
 	int reuse = 1;
 	struct sockaddr_in addr;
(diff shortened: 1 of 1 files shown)


nicolasff:
I cherry-picked this commit as de224445761e9a33dff22e105b8874d89ab8ee1f since this PR had a conflict due to the move of `server.c` under `src/`.

Thanks for adding this! And sorry it took so long to integrate.

Issue #159, opened 2019-05-31
websocket does not handle multiple requests in a packet


granolamatt:
While using chrome and subscribing to multiple (64 was my case) keys in a for loop, webdis will drop requests when the requests are grouped into a single network packet.  I assume that the grouping is because of Nagle's algorithm.

If I change the if statement to a while statement in websocket.c function ws_add_data it fixes my problem.  I am submitting a pull request as well for this issue.

nicolasff:
Let's close this one to have a single issue tracking this bug.

PR #160, opened 2019-05-31
Fixed websocket not reading multiple messages in one packet

granolamatt:
While using chrome and subscribing to multiple (64 was my case) keys in a for loop, webdis will drop requests when the requests are grouped into a single network packet. I assume that the grouping is because of Nagle's algorithm.

This pull request is my proposal to fix issue #159


nicolasff:
Thanks for the pull request!
Merged and tagged as [webdis-0.1.6](https://github.com/nicolasff/webdis/releases/tag/0.1.6).

PR #166, opened 2019-10-29
build: optimized Docker configuration

mickael-palma-argus:
Use multi-stage build.
From 278MB to 99.2MB image size.
Add docker-entrypoint script for dynamic setup.
Add docker-compose file.


nicolasff:
The new Alpine-based docker image also uses a multi-stage build, but brings the size down to 9.5 MB:
```
$ docker images
REPOSITORY                                 TAG                 IMAGE ID            CREATED             SIZE
webdis                                     latest              f384be97f47b        2 days ago          9.5MB
```

PR #169, opened 2020-02-21
Update Dockerfile to use Alpine Linux

jessie-murray:
* Change base image to Alpine 3.11.3
* Use multi-stage build (reducing size from 276 MB to 9.5 MB)
* Change Makefile to build with -O3 instead of -O0 -ggdb


nicolasff:
Hi @jessie-murray, thanks for the PR! I didn't expect multi-stage builds to make such a difference, this is great.

PR #172, opened 2020-03-16
Do not add slash between between DESTDIR and PREFIX in makefile

marsam:
This change will allow to only use PREFIX and leave DESTDIR empty.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,8 @@ OBJS=src/webdis.o src/cmd.o src/worker.o src/slog.o src/server.o src/acl.o src/m
 PREFIX ?= /usr/local
 CONFDIR ?= $(DESTDIR)/etc
 
-INSTALL_DIRS = $(DESTDIR) \
-	       $(DESTDIR)/$(PREFIX) \
-	       $(DESTDIR)/$(PREFIX)/bin \
+INSTALL_DIRS = $(DESTDIR)$(PREFIX) \
+	       $(DESTDIR)$(PREFIX)/bin \
 	       $(CONFDIR)
 
 all: $(OUT) Makefile
@@ -57,7 +56,7 @@ clean:
 	rm -f $(OBJS) $(OUT) $(OBJS_DEPS)
 
 install: $(OUT) $(INSTALL_DIRS)
-	cp $(OUT) $(DESTDIR)/$(PREFIX)/bin
+	cp $(OUT) $(DESTDIR)$(PREFIX)/bin
 	cp webdis.prod.json $(CONFDIR)
 
 


nicolasff:
Hi @marsam, thanks for the PR!
I'm not entirely sure I follow what this allows that didn't work before.
If you wanted to install it to a different location, couldn't you just use e.g. this?

```DESTDIR=~/webdis PREFIX= make install```

This does create `~/webdis/bin/webdis` and `~/webdis/etc/webdis.prod.json`.

I understand that with this change we could swap `DESTDIR` and `PREFIX`, but the purpose of `DESTDIR` is to set the destination directory and it seems to serve this purpose just fine right now.

**Comment by nicolasff** [comment]
@marsam any update on the questions above?

**Comment by marsam** [comment]
Hi, thanks for answering, and sorry for being vague in the PR description.

>I understand that the idea of this change is that we could swap DESTDIR and PREFIX, but the purpose of DESTDIR is to set the destination directory and it seems to serve this purpose just fine right now.

[`DESTDIR`](https://www.gnu.org/software/automake/manual/html_node/DESTDIR.html) is used as poor man's sandbox. In [nixpkgs](https://github.com/NixOS/nixpkgs) we usually refrain from using `DESTDIR`, because it might have _negative_ side-effects https://github.com/NixOS/nixpkgs/issues/65718.
I'm aware that webdis is not really affected by those, since it only installs the binary, but it would be nice if we (nixpkgs) could not have to use DESTDIR for consistence' sake.

Removing the extra slash (`/`) should not cause any regression to your setup, for instance 

    DESTDIR=~/webdis-install PREFIX= make install
should still work, but with leaving `DESTDIR` empty you would need to set `CONFDIR` for `make install`.

I hope it sounds reasonable.

P.S. Thanks for writing webdis, It's really useful

**Comment by nicolasff** [comment]
With your change, running the command you just quoted now prints a warning, which was not there before the change:

```
$ DESTDIR=~/webdis-install PREFIX= make install
-----> Makefile:53: target '/home/testuser/webdis-install' given more than once in the same rule
mkdir -p /home/testuser/webdis-install
mkdir -p /home/testuser/webdis-install/bin
mkdir -p /home/testuser/webdis-install/etc
cp webdis /home/testuser/webdis-install/bin
cp webdis.prod.json /home/testuser/webdis-install/etc
```

marsam:
fixed

nicolasff:
Great, now both of these work as expected:
```DESTDIR= CONFDIR=~/webdis-install/etc PREFIX=~/webdis-install make install```
and
```DESTDIR=~/webdis-install PREFIX= make install```

Thanks again!

PR #175, opened 2020-05-30
Support environment variables in the config

jessie-murray:
* Adds support for environment variables in webdis.json
* Values starting with a `$` and in all caps are interpreted as
environment variables

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -16,6 +16,49 @@
 static struct acl *
 conf_parse_acls(json_t *jtab);
 
+int
+conf_str_allcaps(const char *s, const size_t sz) {
+	size_t i;
+	for(i = 0; i < sz; i++) {
+		if(s[i] != toupper(s[i])) {
+			return 0;
+		}
+	}
+	return 1;
+}
+
+char *
+conf_string_or_envvar(const char *val) {
+	if(!val) {
+		return strdup(""); /* safe value for atoi/aol */
+	}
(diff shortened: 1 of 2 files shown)


src/conf.c
```diff
@@ -52,51 +88,71 @@ conf_read(const char *filename) {
...
+			free(daemonize);
 		} else if(strcmp(json_object_iter_key(kv),"pidfile") == 0 && json_typeof(jtmp) == JSON_STRING){
-			conf->pidfile = strdup(json_string_value(jtmp));
+			conf->pidfile = conf_string_or_envvar(json_string_value(jtmp));
 		} else if(strcmp(json_object_iter_key(kv), "websockets") == 0 && json_typeof(jtmp) == JSON_TRUE) {
 			conf->websockets = 1;
+		} else if(strcmp(json_object_iter_key(kv), "websockets") == 0 && json_typeof(jtmp) == JSON_STRING) {
+			char *ws = conf_string_or_envvar(json_string_value(jtmp));
+			conf->websockets = ws && strcasecmp(ws, "true") == 0 ? 1 : 0;
+			free(ws);
```
nicolasff, line 145: Just like for `atoi_free`, this block and the one for `daemonize` above could be refactored using a single `is_true_free` function, removing the two temporary variables and simplifying the code.

jessie-murray:
Thanks for reviewing, I updated `conf.c` as requested and added a section in `README.md`.

Issue #176, opened 2020-06-16
Webdis do not supports redis replies with the nested arrays.


majklik:
Webdis actually do not supports replies which uses nested arrays. The array element is converted into NULL.
I wrote small path which solve this for JSON output. It tioo handle specific output of the XRANGE/XREAD commands like older implementation for HGETALL:
```diff
--- src/formats/json.c.orig	2020-06-15 20:54:06.000000000 +0100
+++ src/formats/json.c	2020-06-16 09:00:45.000000000 +0100
@@ -89,9 +89,12 @@
 }
 
 static json_t *
-json_hgetall_reply(const redisReply *r) {
+json_exand_array(const redisReply *r);
+
+static json_t *
+json_keyvalue_reply(const redisReply *r) {
 	/* zip keys and values together in a json object */
-	json_t *jroot;
+	json_t *jroot, *jlist;
 	unsigned int i;
 
 	if(r->elements % 2 != 0) {
@@ -103,21 +106,113 @@
 		redisReply *k = r->element[i], *v = r->element[i+1];
 
 		/* keys and values need to be strings */
-		if(k->type != REDIS_REPLY_STRING || v->type != REDIS_REPLY_STRING) {
+		if(k->type != REDIS_REPLY_STRING) {
 			json_decref(jroot);
 			return NULL;
 		}
-		json_object_set_new(jroot, k->str, json_string(v->str));
+		switch (v->type) {
+		case REDIS_REPLY_STRING:
+			json_object_set_new(jroot, k->str, json_string(v->str));
+			break;
+		case REDIS_REPLY_INTEGER:
+			json_object_set_new(jroot, k->str, json_integer(v->integer));
+			break;
+		case REDIS_REPLY_ARRAY:
+			if (!(jlist=json_exand_array(v)))
+				jlist = json_null();
+
+			json_object_set_new(jroot, k->str, jlist);
+			break;
+		default:
+			json_decref(jroot);
+			return NULL;
+		} 
+		
 	}
 	return jroot;
 }
 
 static json_t *
-json_wrap_redis_reply(const struct cmd *cmd, const redisReply *r) {
+json_exand_array(const redisReply *r) {
+
+	unsigned int i;
+	json_t *jlist, *sublist;
+	redisReply *e;
+
+	jlist = json_array();
+	for(i = 0; i < r->elements; ++i) {
+		e = r->element[i];
+		switch(e->type) {
+		case REDIS_REPLY_STRING:
[...]

PR #177, opened 2020-06-16
JSON nested arrays support

majklik:
Add support for nested arrays replies in JSON formatter and some commands outputs are converted to map list in similar way how HGETALL ( XRANGE / XREVRANGE / XCLAIM / XPENDING / XREAD* / GEORADIUS* ).

### DIFF
--- a/src/formats/json.c
+++ b/src/formats/json.c
@@ -181,7 +181,7 @@ json_singlestream_list(const redisReply *r) {
 	const redisReply *id, *msg;
 	const redisReply *e;
 
-	/* reply on XRANGE / XREVRAGE / XCLAIM and one substream from XREAD / XREADGROUP */ 
+	/* reply on XRANGE / XREVRANGE / XCLAIM and one substream of XREAD / XREADGROUP */ 
 	jlist = json_array();
 	for(i = 0; i < r->elements; i++) {
 		e=r->element[i];
@@ -207,7 +207,7 @@ json_xreadstream_list(const redisReply *r) {
 	const redisReply *sid, *msglist;
 	const redisReply *e;
 
-	/* reply on XPEAD / XREADGROUP */ 
+	/* reply on XREAD / XREADGROUP */ 
 	if(r->elements) jobj = json_object();
 	for(i = 0; i < r->elements; i++) {
 		e=r->element[i];
--- a/src/formats/json.c
+++ b/src/formats/json.c
@@ -89,9 +89,12 @@ json_info_reply(const char *s) {
 }
 
 static json_t *
-json_hgetall_reply(const redisReply *r) {
+json_exand_array(const redisReply *r);
+
+static json_t *
+json_array_to_keyvalue_reply(const redisReply *r) {
 	/* zip keys and values together in a json object */
-	json_t *jroot;
+	json_t *jroot, *jlist;
 	unsigned int i;
 
 	if(r->elements % 2 != 0) {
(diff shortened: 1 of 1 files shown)


nicolasff:
Awesome, thanks! I'll go over it shortly.

majklik:
Well, we can continue there with issue https://github.com/nicolasff/webdis/issues/176#issue-639463065 .
The original diff was dirty hack for fast work and supports only two-three commands which I need.

A added in this PR support for other nested arrays replies which is usefull to convert imto map list (IMHO). :-)

PS: We are using webdis about 7 years. :-)

nicolasff:
Hi @majklik,

First off, congratulations on this great pull request. I see in your GitHub profile that this is the first one you've ever opened from this account, and few (if any) contributions to Webdis have been this complex. The code was clear, well-written and easy to follow. Great job!

Going over it a few times, the only real changes I felt were needed we stylistic, in order to maintain a consistent style throughout the code base. None of them actually change any logic, order, or variable names. They were pretty much all about spacing and braces.
I also reviewed the Redis documentation for these commands, and tried to ensure that the keys returned by Webdis as it constructs all of these nested objects stay consistent with the naming conventions used throughout the docs. They are listed below, you will see that they are minor.

Formatting:
* Added `{}` after each single-line `if`. See [goto fail](https://www.imperialviolet.org/2014/02/22/applebug.html) for the reason.
* Added spaces around operators, e.g. `e=r->element[i];` became `e = r->element[i];`.

Actual changes:
* Changed `msgsperowner` to `msgsperconsumer`, following the documentation which says *list every consumer in the consumer group with at least one pending message, and the number of pending messages it has.* – emphasis on *list every consumer*.
* Changed `delivers` to `deliveries`, again following the docs: *The number of times this message was delivered.* "delivers" is not a noun and as such is not well suited for a key in an object.
* In the `GEORADIUS` code, changed `member` to `name`. The docs say: *the first item in the sub-array is always the name of the returned item*.
* Also for `GEORADIUS`, changed `coord` to `coords` since *coordinates* is always plural.

Other: I also renamed `json_exand_array` to `json_expand_array`.


One thing I thought was missing with the `XCLAIM` command, when it's used with `JUSTID`, but after testing it I realized that you had already covered this case with the extra condition `r->element[0]->type == REDIS_REPLY_ARRAY`. Well done!

Thanks again for this great submission, I am very grateful for it.

nicolasff:
An addition note on testing: I used a custom Dockerfile to run Webdis in valgrind and check for leaks (drop this in the Webdis directory):
```dockerfile
FROM debian:buster

RUN apt-get update && apt-get upgrade
RUN apt-get install -y clang valgrind make libevent-dev libmsgpack-dev git redis-server
RUN mkdir /tmp/webdis
COPY . /tmp/webdis/
RUN cd /tmp/webdis && find src -name '*.d' -delete && find src -name '*.o' -delete && sed -i 's/-O3/-g -O0/g' Makefile && make clean all install

RUN echo "daemonize yes" >> /etc/redis.conf
RUN sed -i 's/"daemonize":.*true/"daemonize": false/g' /etc/webdis.prod.json
CMD /usr/bin/redis-server /etc/redis.conf >/dev/null 2>/dev/null && valgrind --tool=memcheck --leak-check=full --show-reachable=yes --show-possibly-lost=yes /usr/local/bin/webdis /etc/webdis.prod.json

EXPOSE 7379
```

After which I ran the Python tests and the following streams tests:

```sh
#!/bin/zsh
curl -s -g 'http://localhost:7379/DEL/stream1' | jq
curl -s -g 'http://localhost:7379/XADD/stream1/*/name/Sara/surname/OConnor' | jq
curl -s -g 'http://localhost:7379/XADD/stream1/*/field1/value1/field2/value2/field3/value3' | jq
curl -s -g 'http://localhost:7379/XLEN/stream1' | jq
curl -s -g 'http://localhost:7379/XGROUP/CREATE/stream1/group1/0' | jq
curl -s -g 'http://localhost:7379/XREADGROUP/GROUP/group1/consumer1/COUNT/1/STREAMS/stream1/>' | jq . > xreadgroup.json
jq < xreadgroup.json
local eventid=$(jq '.XREADGROUP.stream1[0].id' < xreadgroup.json | sed 's/"//g')
curl -s -g 'http://localhost:7379/XCLAIM/stream1/group1/consumer1/0/0-0' | jq
curl -s -g 'http://localhost:7379/XCLAIM/stream1/group1/consumer1/0/'$eventid'/JUSTID' | jq
curl -s -g 'http://localhost:7379/XCLAIM/stream1/group1/consumer1/0/'$eventid | jq
```

Which produces this output:
```json
{
  "DEL": 1
}
{
  "XADD": "1592779353712-0"
}
{
  "XADD": "1592779353721-0"
}
{
  "XLEN": 2
}
{
  "XGROUP": [
    true,
    "OK"
  ]
}
{
  "XREADGROUP": {
    "stream1": [
      {
        "id": "1592779353712-0",
        "msg": {
          "name": "Sara",
          "surname": "OConnor"
        }
      }
    ]
  }
}
{
  "XCLAIM": []
}
{
  "XCLAIM": [
    "1592779353712-0"
  ]
}
{
  "XCLAIM": [
    {
      "id": "1592779353712-0",
      "msg": {
        "name": "Sara",
        "surname": "OConnor"
      }
    }
  ]
}
```

The leak summary is as expected: the only retained memory is from `strdup` calls made when we read the config file. These are not true leaks but just memory that is allocated once per process and not free'd on shutdown.
So the new code paths look good and I don't expect any memory issues.

PR #178, opened 2020-08-13
Ensure websockets write all data; Always keep_alive every websocket

shawwn:
Summary:

- Ensure websockets write all data

- Always keep_alive every websocket

Explanation:

I've been using webdis websockets to write large amounts of data to a webgl demo for plotting lots of data points (https://twitter.com/theshawwn/status/1292266967940960257). There were two hurdles:

1. webdis doesn't handle the return value of `write()` properly, leading to truncated sends

2. webdis seems to close websockets after each request.

This PR writes all data, and also marks every websocket as `keep_alive = 1`, ensuring that it stays open for multiple requests.

There is currently a demo of this here: http://test.tensorfork.com/ though I'm not sure how long that link will remain valid.

### DIFF
--- a/src/websocket.c
+++ b/src/websocket.c
@@ -380,9 +380,7 @@ ws_reply(struct cmd *cmd, const char *p, size_t sz) {
 
 	/* send WS frame */
 	r = http_response_init(cmd->w, 0, NULL);
-	if (cmd_is_subscribe(cmd)) {
-		r->keep_alive = 1;
-	}
+	r->keep_alive = 1;
 	
 	if (r == NULL)
 		return -1;
--- a/src/websocket.c
+++ b/src/websocket.c
@@ -77,10 +77,26 @@ ws_compute_handshake(struct http_client *c, char *out, size_t *out_sz) {
 	return 0;
 }
 
+
+void
+ws_write_all(int socket, void *buffer, size_t length) {
+    size_t written = 0;
+    int n;
+    while (written < length) {
+        int n = write(socket, (char*)buffer + written, length - written);
+        if (n <= 0) {
+            if (errno == EINTR || errno == EAGAIN)
+                continue;
+            err(EXIT_FAILURE, "Could not write() data");
+        }
+        written += n;
+    }
+}
+
+
 int
 ws_handshake_reply(struct http_client *c) {
 
-	int ret;
 	char sha1_handshake[40];
 	char *buffer = NULL, *p;
 	const char *origin = NULL, *host = NULL;
@@ -159,8 +175,7 @@ ws_handshake_reply(struct http_client *c) {
 	p += sizeof(template4)-1;
 
 	/* send data to client */
-	ret = write(c->fd, buffer, sz);
-	(void)ret;
+	ws_write_all(c->fd, buffer, sz);
 	free(buffer);
 
 	return 0;


src/websocket.c
```diff
@@ -77,10 +77,26 @@ ws_compute_handshake(struct http_client *c, char *out, size_t *out_sz) {
...
+void
+ws_write_all(int socket, void *buffer, size_t length) {
+    size_t written = 0;
+    int n;
+    while (written < length) {
+        int n = write(socket, (char*)buffer + written, length - written);
+        if (n <= 0) {
+            if (errno == EINTR || errno == EAGAIN)
+                continue;
+            err(EXIT_FAILURE, "Could not write() data");
```
nicolasff, line 90: This should use the logging interface instead of terminating the service with `err`. You can pass a pointer to the `server` struct from the caller and use it like this:

```c
#include "slog.h"

// ...

int
ws_write_all(struct server *s, int fd, void *buffer, size_t length) {

// ...

const char err_msg[] = "Could not write data to websocket";
slog(s, WEBDIS_ERROR, err_msg, sizeof(err_msg)-1);
return -1;
```

Instead of calling `err`, a return code should be used to tell the caller about the failure so that it can cleanly terminate the connection and free the WS structure.
Successful writes should `return 0;` after the `while` loop. See comment below too.

Unfortunately I see that even if this error is propagated from `ws_write_all` to `ws_handshake_reply` to the caller of `ws_handshake_reply`, the return code is unused in `worker.c`. There, it should instead be used to set `c->broken = 1` just like a couple of lines below the call to `ws_handshake_reply`.

Something like this. Instead of:
```c
if(nparsed < ret) {
	http_client_add_to_body(c, c->buffer + nparsed + 1, c->sz - nparsed - 1);
	ws_handshake_reply(c);
}
```

you'd want to use the return value:
```c
if(nparsed < ret) {
	http_client_add_to_body(c, c->buffer + nparsed + 1, c->sz - nparsed - 1);
	if(ws_handshake_reply(c) < 0)
		c->broken = 1;
}
```
I'm also not sure about this repeated write outside of the event loop. Ideally the call to `write(2)` should happen when the file descriptor is writable, and the event re-registered if there's still more data to write. This is a more complex change though, and could be done in a second pass.

src/websocket.c
```diff
@@ -159,8 +175,7 @@ ws_handshake_reply(struct http_client *c) {
 	p += sizeof(template4)-1;
 
 	/* send data to client */
-	ret = write(c->fd, buffer, sz);
-	(void)ret;
+	ws_write_all(c->fd, buffer, sz);
```
nicolasff, line 178: The return value of this function should be propagated to the caller in `worker.c`. Can you put back the `int ret;` and use it?

Also, to pass a `server` struct to be used in `slog`, change to:
```c
ret = ws_write_all(c->s, c->fd, buffer, sz);
```

(also see comment below)

src/websocket.c
```diff
@@ -159,8 +175,7 @@ ws_handshake_reply(struct http_client *c) {
...
 	free(buffer);
 
 	return 0;
```
nicolasff, line 181: Here we'd use `ret`:
```c
return ret;
```

shawwn:
Okay, I'll look into this when I have some time. Your comments were thorough and helpful; thank you.

> In particular, terminating the service because a websocket client got disconnected is (slightly) overkill 😊.

It doesn't. In fact, according to the `write` manpage (https://www.man7.org/linux/man-pages/man2/write.2.html), there don't seem to be any error codes that can cause the `err` to fire. Most error codes are related to file system errors, for example.

It's a lot more work to handle the error code gracefully. Are you sure there are situations where `write` can fail in a way that triggers the `err`?

I agree it's probably better to handle it, but FWIW I've been running this code for a week now with no problems at http://test.tensorfork.com/.

> I'm also not sure about this repeated write outside of the event loop. Ideally the call to write(2) should happen when the file descriptor is writable, and the event re-registered if there's still more data to write. This is a more complex change though, and could be done in a second pass.

Yeah, the repeated write is concerning. It opens up the server to a "slow attack", where a client only downloads a few bytes per second. What are some solutions? We could spawn a thread for each connected websocket.

PR #179, opened 2020-08-19
CodeQL: apply recommendations

jessie-murray:
This commit corrects some of the warnings reported by CodeQL: https://lgtm.com/projects/g/nicolasff/webdis/?mode=list

* Use `localtime_r` instead of `localtime`
* Use correct argument type in callback to `msgpack_packer_new`
* Address FIXME in `conf.c`
* Remove redundant check in `websocket.c`

I didn't update the source files that come from third-party dependencies like jansson and hiredis, maybe it makes more sense to update them to more recent versions?

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -237,7 +237,11 @@ conf_parse_acl(json_t *j) {
 		a->http_basic_auth = calloc(len, 1);
 		
 		base64_init_encodestate(&b64);
-		pos = base64_encode_block(plain, (int)plain_len, a->http_basic_auth, &b64); /* FIXME: check return value */
+		pos = base64_encode_block(plain, (int)plain_len, a->http_basic_auth, &b64);
+		if(!pos) { /* nothing was encoded */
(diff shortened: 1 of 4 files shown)


nicolasff:
Hi Jesse!

Thanks for another PR! I'll go over it shortly. I assume you saw my recent attempts at enabling CodeQL through GitHub actions. I just couldn't get it to work, I even used [Act](https://github.com/nektos/act) to run GitHub Actions locally and had similar issues: CodeQL would just fail to start and not give any error message or indication of what went wrong. I went through all of the docs, tried all the different options to run it, but it never even made any progress so I think I'll give up on it for now.

Yours still look like valuable changes to include.

As for your question about dependencies, I was wondering about these recently since they haven't been touched in years. I remember that there are quite a few warnings coming from hiredis, so that might be a good one to try to update first. I used to work on [phpredis](https://github.com/phpredis/phpredis) with one of the guys who now maintains both, so I could reach out to him if there are still issues.

Thanks again, and I'll make sure this doesn't stay open for long.

PR #186, opened 2021-03-12
Warning fixes

jessie-murray:
I fixed a few compilation warnings: 2 in `conf.c` from the recent changes I made there, and also in `dict.c` where some functions are unused and cause a warning. I noticed also that hiredis was from an old import, maybe it should be upgraded?

1. Remove unused size_t sz variable
2. Remove const on free()'d variable
3. Remove unused function warning in dict.c

### DIFF
--- a/src/b64/cencode.c
+++ b/src/b64/cencode.c
@@ -48,6 +48,7 @@ int base64_encode_block(const char* plaintext_in, int length_in, char* code_out,
 			result = (fragment & 0x0fc) >> 2;
 			*codechar++ = base64_encode_value(result);
 			result = (fragment & 0x003) << 4;
+			/* fall through */
 	case step_B:
 			if (plainchar == plaintextend)
 			{
@@ -59,6 +60,7 @@ int base64_encode_block(const char* plaintext_in, int length_in, char* code_out,
 			result |= (fragment & 0x0f0) >> 4;
 			*codechar++ = base64_encode_value(result);
 			result = (fragment & 0x00f) << 2;
+			/* fall through */
 	case step_C:
 			if (plainchar == plaintextend)
 			{
--- a/src/conf.c
+++ b/src/conf.c
@@ -199,7 +199,6 @@ acl_read_commands(json_t *jlist, struct acl_commands *ac) {
 	for(i = 0, cur = 0; i < json_array_size(jlist); ++i) {
 		json_t *jelem = json_array_get(jlist, i);
 		if(json_typeof(jelem) == JSON_STRING) {
-			size_t sz;
 			ac->commands[cur] = conf_string_or_envvar(json_string_value(jelem));
 			cur++;
 		}
@@ -241,7 +240,7 @@ conf_parse_acl(json_t *j) {
 		base64_encodestate b64;
 		int pos;
 		char *p;
-		const char *plain = conf_string_or_envvar(json_string_value(jbasic));
+		char *plain = conf_string_or_envvar(json_string_value(jbasic));
 		size_t len, plain_len = strlen(plain) + 0;
 		len = (plain_len + 8) * 8 / 6;
 		a->http_basic_auth = calloc(len, 1);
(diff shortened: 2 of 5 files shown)


nicolasff:
Thanks @jessie-murray!

[The CircleCI build](https://app.circleci.com/pipelines/github/nicolasff/webdis/91/workflows/47fe37a7-8ae7-49bc-a039-d75b243e7b79/jobs/87/parallel-runs/0/steps/0-104) shows another one that might be good to add. I don't know if you have access to the output, so here it is again:
```
src/formats/msgpack.c: In function 'msgpack_wrap_redis_reply':
src/formats/msgpack.c:170:2: warning: passing argument 2 of 'msgpack_packer_new' from incompatible pointer type [enabled by default]
  msgpack_packer* pk = msgpack_packer_new(out, on_msgpack_write);
  ^
In file included from /usr/include/msgpack.h:25:0,
                 from src/formats/msgpack.h:4,
                 from src/formats/msgpack.c:1:
/usr/include/msgpack/pack.h:124:24: note: expected 'msgpack_packer_write' but argument is of type 'int (*)(void *, const char *, size_t)'
 inline msgpack_packer* msgpack_packer_new(void* data, msgpack_packer_write callback)
                        ^
```

You may not have seen it if you're building Webdis on a machine without msgpack installed, since the Makefile detects it and adds `-DMSGPACK=1` if the library is present.

nicolasff:
TIL about `-Wimplicit-fallthrough`!
```
-Wimplicit-fallthrough=0 disables the warning altogether.
-Wimplicit-fallthrough=1 matches .* regular expression, any comment is used as fallthrough comment.
-Wimplicit-fallthrough=2 case insensitively matches .*falls?[ \t-]*thr(ough|u).* regular expression.
```
[Values of 3 or more](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) use even more complex regular expressions 🤨

nicolasff:
Looks great as always, thanks!

nicolasff:
@jessie-murray oh and regarding your earlier comment about hiredis: I had tried to upgrade it [in the past](https://github.com/nicolasff/webdis/tree/warning-fixes) and had some weird issues so I had given up. You're welcome to give it a shot but it was not trivial from what I remember.

jessie-murray:
@nicolasff I looked into upgrading hiredis yesterday with the thought that maybe a more recent version would be faster than the code that's currently checked in. I couldn't see any measurable difference in the request rate or latency, and the CPU usage was also similar in both versions. I didn't compare memory usage, now that I think about it it might be different.

I ran a modified version of the `ab` command taken from `bench.sh`, just with more clients and more requests to create sustained load over a longer time span. The difference was something like 91,600 requests per second before the change and 91,400 after the change which I interpreted as being the same. That said when I started the test these numbers were much lower with maybe 19,000 requests per second only, and it's only when I reduced the log level to zero that it became faster. I noticed that Webdis calls `fsync` after each log line, and I think this is probably what is causing the performance to drop. Maybe the `fsync` could be removed or made optional since this is just a log? Another way to improve performance would be to buffer log events and have a thread flushing them in batches, but this would require a much larger change. I'll look into making the `fsync` optional to see what kind of difference it makes.

nicolasff:
@jessie-murray oh wow ok… I didn't remember this being the case, I'll check out the difference it makes. As for a change to change the fsync behavior, you're more than welcome to submit a PR since I won't be able to work on it. That said, I'm not convinced that adding a buffering system should be the first change; it seems like it would need a lot of code with all the potential for bugs that comes with this kind of rewrite of the logging system – which is what it feels like. This kind of change needs to demonstrate a clear advantage due to its high risk of introducing bugs.

Before exploring the buffering option I would suggest looking at the way [Cassandra](https://cassandra.apache.org/doc/latest/architecture/storage_engine.html) configures its fsync frequency with `commitlog_sync`. This is for actual data written to this database though, so I'm not sure whether it would make sense to have the same options for "just" log events but it's a pretty flexible approach which might work here.

PR #191, opened 2021-03-28
Add build (+ test) workflow for GitHub Actions

jessie-murray:
Performs the same tasks as on CircleCI and dds badge to README.md.
No change made to the CircleCI config.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,32 @@
+name: Build
+
+# trigger the workflow on push or pull requests
+on: [push, pull_request]
+
+jobs:
+  build-and-run-tests:
+    runs-on: ubuntu-18.04     # this is a GitHub Runner, hosting the execution
+    container: ubuntu:18.04   # but this is a Docker Hub container, in which everything runs
+
+    services:
+      redis:
+        image: redis:6.2.1-alpine  # Docker Hub image used as a sidecar
+
+    steps:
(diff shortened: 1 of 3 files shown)


nicolasff:
Thanks! That looks great.

Looking at the job names, I was wondering if there was a way to have both the PR job say "build and run tests" but the badge say only "Build"? I think it makes more sense to have the PR check say "build and run tests" but this string next to "passing" doesn't look quite right. If not, then don't worry about it and we can keep "Build" just for the badge.

Also (and it's very minor, I can probably fix that on merge) there's a typo in the commit message: s/dds/adds.

nicolasff (changes requested):
Thanks! That looks great.

Looking at the job names, I was wondering if there was a way to have both the PR job say "build and run tests" but the badge say only "Build"? I think it makes more sense to have the PR check say "build and run tests" but this string next to "passing" doesn't look quite right. If not, then don't worry about it and we can keep "Build" just for the badge.

Also (and it's very minor, I can probably fix that on merge) there's a typo in the commit message: s/dds/adds.

jessie-murray:
I can't find a way to split it properly, but I've updated it now to use "Build" for the workflow name (on the badge) with also "build and run tests" for the job name, which is listed on this PR. Is that a good compromise?

I also switched to the Alpine version of Redis (32 MB vs 105 MB).

PR #192, opened 2021-04-10
CodeQL fixes

jessie-murray:
Some small fixes based on the CodeQL warnings.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -16,6 +16,7 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/ioctl.h>
+#include <limits.h>
 
 /**
  * Sets up a non-blocking socket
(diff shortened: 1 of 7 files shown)


src/server.c
```diff
@@ -163,10 +164,17 @@ server_daemonize(const char *pidfile) {
...
+			char pid_buffer[(CHAR_BIT * sizeof(int) / 3) + 3]; /* max length for int */
+			int pid_sz = snprintf(pid_buffer, sizeof(pid_buffer), "%d\n", (int)getpid());
+			size_t written;
+			int written_total = 0;
+			while ((written = write(pid_fd, pid_buffer + written_total, pid_sz - written_total)) > 0
+				&& written_total < pid_sz) {
+				written_total += written;
+			}
+			close(pid_fd);
 		}
```
nicolasff, line 178: we should log if we can't write the pid. Also, `write(2)` returns `ssize_t` (not `size_t`) and I'd suggest adding `O_TRUNC` and `O_NOFOLLOW`.

nicolasff (changes requested):
Nice! Thanks, that's a good start. It's also good to see that nothing major was highlighted, these are relatively harmless warnings.

PR #195, opened 2021-05-09
Only process `Connection: close` header if full request was read (fixes #194)

### LINKED ISSUE #194: Connection: close triggers a freeze for payloads bigger than certain limit
Thanks for webdis, I like it a lot. I spotted the following issue when I put webdis behind a nginx, which I use for reverse-proxying:

If I send a PUT with some small payload, then webdis is fine, even when the `Connection: close` header is present:
```shell
(echo "[1"; yes ",1" | head -n 1957; echo "]" )  | curl -XPUT http://172.17.0.2:7379/SET/toobig -d @- -H "Connection: close"
```
but if the payload is bigger, then the connection hangs and eventually times out:
```shell
(echo "[1"; yes ",1" | head -n 1958; echo "]" )  | curl -XPUT http://172.17.0.2:7379/SET/toobig -d @- -H "Connection: close"
```
finally the problem goes away if I remove the `Connection: close` header.
```shell
(echo "[1"; yes ",1" | head -n 1958; echo "]" )  | curl -XPUT http://172.17.0.2:7379/SET/toobig -d @-
```

This is a problem, because the default configuration of nginx does append the `Connection: close`.
I worked that around by instructing nginx to send `Connection: keep-alive` instead:
```
proxy_set_header Connection keep-alive;
```

but I think that this is better addressed at webdis directly. For start it might be good to add that to webdis known issues in the docs.

jessie-murray:
The presence of a `Connection: close` header sets a flag on the request,
which we process by no longer running the HTTP parser as soon as it
returns from parsing a chunk. Since HTTP 100 uploads require at least 2
reads, we need to handle the `Connection: close` only if we've reached
the end of the request. This change sets a flag in `on_message_complete`,
which we use in combination with the header to stop reading.

Also adding a test to reproduce this issue and including it in CI job.

@nicolasff I tried to find an existing flag but didn't find anything already set in `http_client_on_message_complete` that I could use. There's `c->keep_alive = 0` just above the flag I set, but it's 0 by default due to the `calloc` so I don't think it's a reliable indicator that the request has been fully parsed. I was also thinking there might be something inside the parser object itself but I didn't really want to rely on its internals (what if it changes later?). But if you have a better idea it might avoid this 1-byte increase, even though this is very small and might actually not make any difference due to alignment.w

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Install dependencies
       run: |
         apt-get -y update
-        apt-get -y --force-yes install make gcc libevent-dev libmsgpack-dev python3
+        apt-get -y --force-yes install make gcc libevent-dev libmsgpack-dev python3 curl
     - name: Build
       run: make
     - name: Run Webdis and test
@@ -25,6 +25,7 @@ jobs:
         ./webdis .github/workflows/webdis-ci.json
         sleep 2
         ./tests/basic.py
+        ./tests/curl-tests.sh
     - name: Archive logs
       uses: actions/upload-artifact@v2
       with:
--- a/src/client.c
+++ b/src/client.c
@@ -182,6 +182,7 @@ http_client_on_message_complete(struct http_parser *p) {
 	/* keep-alive detection */
 	if (c->parser.flags & F_CONNECTION_CLOSE) {
 		c->keep_alive = 0;
+		c->fully_read = 1; /* only *now* can we stop waiting for input */
 	} else if(c->parser.http_major == 1 && c->parser.http_minor == 1) { /* 1.1 */
 		c->keep_alive = 1;
 	}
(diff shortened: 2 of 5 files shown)


jessie-murray:
[force-push](https://github.com/nicolasff/webdis/compare/6efaa6f2be37184251bcde5ec8a481b6997db85c..3c67055bf8030b557500bc47cc9d06b4cd02e18b) is adding `curl` to the CI job due to:
```
./tests/curl-tests.sh: line 7: curl: command not found
Error: Process completed with exit code 127.
```

nicolasff:
Nice! Thanks for the PR. I looked at `http_parser` and couldn't find a field that indicates that parsing has ended; the callback is triggered from multiple places and it seems better to set a flag explicitly.

On a related topic, I filed #196 to look at `http_parser` which hasn't been touched in years.

PR #198, opened 2021-07-17
Websocket test fixes

jessie-murray:
This is a long PR attempting to resolve the issues with Websocket tests, as we've discussed in the past.
It is best read commit by commit, rather than as a single diff. I started by cleaning up the websocket test client and ended up rewriting a large part of it to use [libevent buffers](https://libevent.org/doc/bufferevent_8h.html), which made it possible for me to actually understand what was happening and to manage the read and write buffers in the event-driven context.

I then focused on the webdis side of websockets, by adding trace logs everywhere in `src/websocket.c`. With the current implementation I don't think they should cause performance problems. These helped me during debugging but now I am less sure of whether they are really needed or not.

While I was fixing this code I also noticed that the HTML page for testing websockets did not have error handling or much styling so I added some CSS and cleaned up the code, it looks prettier now:

image

One thing I was not too sure about was whether the websocket test should be included in the GitHub Action that runs on each pull request, but since this is a lot of new test code maybe it should't be part of the validation yet.

@nicolasff this is my largest change made to webdis so far, I hope it's not too long to review. If this is a problem it could probably be split into several smaller pull requests.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -2,6 +2,7 @@
 #include "server.h"
 #include "worker.h"
 #include "client.h"
+#include "slog.h"
 
 #include <string.h>
 #include <stdlib.h>
@@ -98,7 +99,12 @@ http_response_cleanup(struct http_response *r, int fd, int success) {
 	free(r->out);
 	if(!r->keep_alive || !success) {
 		/* Close fd is client doesn't support Keep-Alive. */
-		fprintf(stderr, "http_response_cleanup: keep_alive=%d, success=%d -> closing\n", r->keep_alive, success);
+		if (r->w && slog_enabled(r->w->s, WEBDIS_TRACE)) {
+			char format[] = "http_response_cleanup: keep_alive=%d, success=%d -> closing";
+			char cleanup_msg[sizeof(format)];
+			snprintf(cleanup_msg, sizeof(cleanup_msg), format, r->keep_alive ? 1 : 0, success ? 1: 0);
+			slog(r->w->s, WEBDIS_TRACE, cleanup_msg, 0);
+		}
 		close(fd);
 	}
 
@@ -125,8 +131,19 @@ http_can_write(int fd, short event, void *p) {
 	if(ret > 0)
 		r->sent += ret;
 
-	fprintf(stderr, "http_can_write: ret=%d, r->out_sz=%lu, r->sent=%d\n",
-		ret, r->out_sz, r->sent);
+	if (r->w && slog_enabled(r->w->s, WEBDIS_TRACE)) { /* trace logs */
+		char format[] = "http_can_write: wrote %d, remaining=%lu, total sent=%d";
+		size_t contents_sz = snprintf(NULL, 0, format, ret, r->out_sz - r->sent, r->sent);
+		char *contents_msg = calloc(contents_sz + 1, 1);
+		if (contents_msg) {
+			snprintf(contents_msg, contents_sz + 1, format, ret, r->out_sz - r->sent, r->sent);
+			slog(r->w->s, WEBDIS_TRACE, contents_msg, contents_sz);
+			free(contents_msg);
+		} else {
+			slog(r->w->s, WEBDIS_ERROR, "Failed allocation in http_can_write", 0);
+		}
+	}
+
 	if(ret <= 0 || r->out_sz - r->sent == 0) { /* error or done */
 		http_response_cleanup(r, fd, (int)r->out_sz == r->sent ? 1 : 0);
 	} else { /* reschedule write */
--- a/src/http.c
+++ b/src/http.c
@@ -98,6 +98,7 @@ http_response_cleanup(struct http_response *r, int fd, int success) {
 	free(r->out);
 	if(!r->keep_alive || !success) {
 		/* Close fd is client doesn't support Keep-Alive. */
+		fprintf(stderr, "http_response_cleanup: keep_alive=%d, success=%d -> closing\n", r->keep_alive, success);
 		close(fd);
 	}
 
@@ -124,6 +125,8 @@ http_can_write(int fd, short event, void *p) {
 	if(ret > 0)
 		r->sent += ret;
 
+	fprintf(stderr, "http_can_write: ret=%d, r->out_sz=%lu, r->sent=%d\n",
+		ret, r->out_sz, r->sent);
 	if(ret <= 0 || r->out_sz - r->sent == 0) { /* error or done */
 		http_response_cleanup(r, fd, (int)r->out_sz == r->sent ? 1 : 0);
 	} else { /* reschedule write */
--- a/src/http.c
+++ b/src/http.c
@@ -2,7 +2,6 @@
 #include "server.h"
 #include "worker.h"
 #include "client.h"
-#include "slog.h"
 
 #include <string.h>
 #include <stdlib.h>
@@ -99,12 +98,6 @@ http_response_cleanup(struct http_response *r, int fd, int success) {
 	free(r->out);
 	if(!r->keep_alive || !success) {
 		/* Close fd is client doesn't support Keep-Alive. */
-		if (r->w && slog_enabled(r->w->s, WEBDIS_TRACE)) {
-			char format[] = "http_response_cleanup: keep_alive=%d, success=%d -> closing";
-			char cleanup_msg[sizeof(format)];
-			snprintf(cleanup_msg, sizeof(cleanup_msg), format, r->keep_alive ? 1 : 0, success ? 1: 0);
-			slog(r->w->s, WEBDIS_TRACE, cleanup_msg, 0);
-		}
 		close(fd);
 	}
 
@@ -130,19 +123,6 @@ http_can_write(int fd, short event, void *p) {
 
 	if(ret > 0)
 		r->sent += ret;
-
-	if (r->w && slog_enabled(r->w->s, WEBDIS_TRACE)) { /* trace logs */
-		char format[] = "http_can_write: wrote %d, remaining=%lu, total sent=%d";
-		size_t contents_sz = snprintf(NULL, 0, format, ret, r->out_sz - r->sent, r->sent);
-		char *contents_msg = calloc(contents_sz + 1, 1);
-		if (contents_msg) {
-			snprintf(contents_msg, contents_sz + 1, format, ret, r->out_sz - r->sent, r->sent);
-			slog(r->w->s, WEBDIS_TRACE, contents_msg, contents_sz);
-			free(contents_msg);
-		} else {
-			slog(r->w->s, WEBDIS_ERROR, "Failed allocation in http_can_write", 0);
-		}
-	}
 
 	if(ret <= 0 || r->out_sz - r->sent == 0) { /* error or done */
 		http_response_cleanup(r, fd, (int)r->out_sz == r->sent ? 1 : 0);
--- a/src/http.c
+++ b/src/http.c
@@ -2,6 +2,7 @@
 #include "server.h"
 #include "worker.h"
 #include "client.h"
+#include "slog.h"
 
 #include <string.h>
 #include <stdlib.h>
@@ -137,7 +138,11 @@ http_schedule_write(int fd, struct http_response *r) {
 	if(r->w) { /* async */
 		event_set(&r->ev, fd, EV_WRITE, http_can_write, r);
 		event_base_set(r->w->base, &r->ev);
-		event_add(&r->ev, NULL);
+		int ret = event_add(&r->ev, NULL);
+		if (ret != 0) { /* could not schedule write */
+			slog(r->w->s, WEBDIS_ERROR, "Could not schedule HTTP write", 0);
+			http_response_cleanup(r, fd, 0);
+		}
 	} else { /* blocking */
 		http_can_write(fd, 0, r);
 	}
--- a/src/slog.c
+++ b/src/slog.c
@@ -77,11 +77,21 @@ slog_fsync_init(struct server *s) {
 }
 
 /**
+ * Returns whether this log level is enabled.
+ */
+int
+slog_enabled(struct server *s, log_level level) {
+	return level <= s->cfg->verbosity ? 1 : 0;
+}
+
+/**
  * Write log message to disk, or stderr.
  */
 void
 slog(struct server *s, log_level level,
 		const char *body, size_t sz) {
+
+	if(level > s->cfg->verbosity) return; /* too verbose */
 
 	const char *c = "EWNID";
 	time_t now;
@@ -90,8 +100,6 @@ slog(struct server *s, log_level level,
 	char msg[124];
 	char line[256]; /* bounds are checked. */
 	int line_sz, ret;
-
-	if(level > s->cfg->verbosity) return; /* too verbose */
 
 	if(!s->log.fd) return;
 
--- a/src/slog.c
+++ b/src/slog.c
@@ -87,13 +87,11 @@ slog_enabled(struct server *s, log_level level) {
 /**
  * Write log message to disk, or stderr.
  */
-void
-slog(struct server *s, log_level level,
+static void
+slog_internal(struct server *s, log_level level,
 		const char *body, size_t sz) {
 
-	if(level > s->cfg->verbosity) return; /* too verbose */
-
-	const char *c = "EWNID";
+	const char *c = "EWNIDT";
 	time_t now;
 	struct tm now_tm, *lt_ret;
 	char time_buf[64];
@@ -128,4 +126,15 @@ slog(struct server *s, log_level level,
 	}
 
 	(void)ret;
+}
+
+/**
+ * This wrapper around slog_internal that first checks the log level.
+ */
+void
+slog(struct server *s, log_level level,
+		const char *body, size_t sz) {
+	if(level <= s->cfg->verbosity) { /* check log level first */
+		slog_internal(s, level, body, sz);
+	}
 }
--- a/src/slog.c
+++ b/src/slog.c
@@ -130,7 +130,7 @@ slog_internal(struct server *s, log_level level,
 }
 
 /**
- * This wrapper around slog_internal that first checks the log level.
+ * Thin wrapper around slog_internal that first checks the log level.
  */
 void
 slog(struct server *s, log_level level,
--- a/src/slog.c
+++ b/src/slog.c
@@ -116,8 +116,9 @@ slog_internal(struct server *s, log_level level,
 	}
 
 	/* generate output line. */
+	char letter = (level == WEBDIS_TRACE ? 5 : c[level]);
 	line_sz = snprintf(line, sizeof(line),
-		"[%d] %s %c %s\n", (int)s->log.self, time_buf, c[level], msg);
+		"[%d] %s %c %s\n", (int)s->log.self, time_buf, letter, msg);
 
 	/* write to log and maybe flush to disk. */
 	ret = write(s->log.fd, line, line_sz);
--- a/src/websocket.c
+++ b/src/websocket.c
@@ -6,6 +6,7 @@
 #include "worker.h"
 #include "pool.h"
 #include "http.h"
+#include "slog.h"
 
 /* message parsers */
 #include "formats/json.h"
@@ -219,13 +220,13 @@ ws_execute(struct http_client *c, const char *frame, size_t frame_len) {
 }
 
 static struct ws_msg *
-ws_msg_new() {
-	fprintf(stderr, "------------ NEW -----------\n");
+ws_msg_new(struct server *s) {
+	slog(s, WEBDIS_TRACE, "ws_msg_new", 0);
 	return calloc(1, sizeof(struct ws_msg));
 }
 
 static void
-ws_msg_add(struct ws_msg *m, const char *p, size_t psz, const unsigned char *mask) {
+ws_msg_add(struct server *s, struct ws_msg *m, const char *p, size_t psz, const unsigned char *mask) {
 
 	/* add data to frame */
 	size_t i;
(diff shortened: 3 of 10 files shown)


nicolasff:
Thanks again @jessie-murray! This is pretty long, but I'll try to go over it this weekend.

Just from a quick look, it doesn't seem like there are tests for pub/sub over websockets. If I remember correctly this was really the one feature that had most of the issues; basic websockets worked but the combination of the two would lead to either crashes or hung clients, I don't really remember since it's been so long. Don't get me wrong this already looks like a great start, but this case would need to be covered before the feature can be enabled by default.

src/slog.c
```diff
@@ -91,8 +101,6 @@ slog(struct server *s, log_level level,
 	char line[256]; /* bounds are checked. */
 	int line_sz, ret;
 
-	if(level > s->cfg->verbosity) return; /* too verbose */
-
```
nicolasff, line 95: I'm not sure that moving this makes a difference, the same amount of space is still pre-allocated on the stack.
Adding a check _before_ calling would save a little bit, at least the function prelude (IIRC).

src/http.c
```diff
@@ -98,6 +99,12 @@ http_response_cleanup(struct http_response *r, int fd, int success) {
 	free(r->out);
 	if(!r->keep_alive || !success) {
 		/* Close fd is client doesn't support Keep-Alive. */
+		if (r->w && slog_enabled(r->w->s, WEBDIS_TRACE)) {
```
nicolasff, line 102: why is this checking for `r->w`, and the one below as well?

src/websocket.c
```diff
@@ -365,16 +415,32 @@ ws_reply(struct cmd *cmd, const char *p, size_t sz) {
...
+	if (r == NULL) {
+		free(frame);
+		slog(cmd->w->s, WEBDIS_ERROR, "Failed response allocation in ws_reply", 0);
+		return -1;
 	}
 
-	if (r == NULL)
-		return -1;
+	/* mark as keep alive, otherwise we'll close the connection after the first reply */
+	r->keep_alive = 1;
```
nicolasff, line 425: this seems like the only actual change in this file, right? the rest being logging?

tests/Makefile
```diff
@@ -1,11 +1,11 @@
 OUT=websocket pubsub
-CFLAGS=-O3 -Wall -Wextra
-LDFLAGS=-levent -lpthread -lrt
+CFLAGS=-O0 -g -Wall -Wextra -I../src/http-parser/
+LDFLAGS=-g -levent -lpthread
```
nicolasff, line 3: this is useful for local tests, but ultimately I think we'll want either `-O3` back or the same `DEBUG=1` logic as in the `Makefile` at the root level.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+#include <ctype.h>
+#include <getopt.h>
 
 #include <sys/types.h>
 #include <sys/socket.h>
 
 #include <event.h>
+#include <http_parser.h>
+
+#define DEBUG_LOGS 0
```
nicolasff, line 20: this seems to be unused.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
 
 struct host_info {
-	char *host;
-	short port;
+    char *host;
+    short port;
```
nicolasff, line 24: hmm… the whole repo uses tabs. I do prefer spaces but I'm not sure this is the right place to start changing the indentation.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+    for (char *cur = p; cur < p + sz; cur += 16) {
+        char letters[16] = {0};
+        int limit = (cur + 16) > p + sz ? (sz % 16) : 16;
+        wt->debug("%08lx ", cur - p); /* address */
+        for (int i = 0; i < limit; i++) {
+            wt->debug("%02x ", (unsigned int)(cur[i] & 0xff));
+            letters[i] = isprint(cur[i]) ? cur[i] : '.';
+        }
+        for (int i = limit; i < 16; i++) { /* pad on last line */
+            wt->debug("   ");
```
nicolasff, line 91: 3 spaces for `%02x `? this could use a comment to explain the unusual string length.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+        }
+        wt->debug(" %.*s\n", limit, letters);
+    }
+}
+
+void
+evbuffer_debug_dump(struct worker_thread *wt, struct evbuffer *buffer) {
+    size_t sz = evbuffer_get_length(buffer);
+    char *data = malloc(sz);
```
nicolasff, line 100: this should be null-checked.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
-			wt->msg_target - wt->msg_received, wt->byte_count);
-	}
-	wt->byte_count += sz;
-
-	/* decrement read count, and stop receiving when we reach zero. */
-	wt->msg_received++;
-	if(wt->msg_received == wt->msg_target) {
-		event_base_loopexit(wt->base, NULL);
-	}
+    // printf("process_message\n");
```
nicolasff, line 118: I realize this was already in the code but it can probably be removed now.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+websocket_can_write(int fd, short event, void *ptr) {
+    int ret;
+    struct worker_thread *wt = ptr;
+    wt->debug("%s (wt=%p, fd=%d)\n", __func__, wt, fd);
+
+    if(event != EV_WRITE) {
+        return;
+    }
+    switch (wt->state)
+    {
```
nicolasff, line 148: minor, but this `{` should be on the same line as the `switch` statement.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
-		event_set(&wt->ev_w, fd, EV_WRITE, websocket_write, wt);
-		event_base_set(wt->base, &wt->ev_w);
-		ret = event_add(&wt->ev_w, NULL);
-	}
+websocket_can_write(int fd, short event, void *ptr) {
+    int ret;
+    struct worker_thread *wt = ptr;
+    wt->debug("%s (wt=%p, fd=%d)\n", __func__, wt, fd);
+
+    if(event != EV_WRITE) {
```
nicolasff, line 144: when would this happen?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
-
-				wt->got_header = 1;
-				event_set(&wt->ev_w, fd, EV_WRITE,
-						websocket_write, wt);
-				event_base_set(wt->base, &wt->ev_w);
-				ret = event_add(&wt->ev_w, NULL);
-			}
-		} else {
-			/* we've had the header already, now bufffer data. */
-			evbuffer_add(wt->buffer, packet, ret);
```
nicolasff, line 116: so this _was_ already using libevent buffers?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
-		event_base_loopexit(wt->base, NULL);
-	}
+websocket_can_read(int fd, short event, void *ptr) {
+    int ret;
+
+    struct worker_thread *wt = ptr;
+    wt->debug("%s (wt=%p)\n", __func__, wt);
+
+    if(event != EV_READ) {
+        return;
```
nicolasff, line 190: same question here, when would this happen?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+    }
+
+    /* read message */
+    ret = evbuffer_read(wt->rbuffer, fd, 65536);
+    wt->debug("evbuffer_read() returned %d; wt->state=%d. wt->rbuffer:\n", ret, wt->state);
+    evbuffer_debug_dump(wt, wt->rbuffer);
+    if (ret == 0) {
+        wt->debug("We didn't read anything from the socket...\n");
```
nicolasff, line 198: should this also mark it as broken?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+            process_message(wt, payload_len);
+
+            if (evbuffer_get_length(wt->rbuffer) == 0) { /* consumed everything */
+                if (wt->msg_received < wt->msg_target) { /* let's write again */
+                    wt->debug("our turn to write again\n");
+                    wt->state = WS_RECEIVED_HANDSHAKE;
+                    ws_enqueue_frame(wt);
+                } /* otherwise, we're done */
+                return;
+            } else {
```
nicolasff, line 248: when would there still be data after reading the full payload with `evbuffer_drain`?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+    event_base_set(wt->base, &wt->ev_r);
+    event_add(&wt->ev_r, NULL);
+}
+
+static int
+ws_on_headers_complete(http_parser *p) {
+    struct worker_thread *wt = p->data;
+
+    wt->debug("%s (wt=%p)\n", __func__, wt);
+    // TODO
```
nicolasff, line 283: is something missing in this callback?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+    return 0;
+}
+
+static void
+ws_enqueue_frame_for_command(struct worker_thread *wt, char *cmd, size_t sz) {
+    unsigned char mask[4];
+    for (int i = 0; i < 4; i++) {
+        mask[i] = rand() & 0xff;
+    }
```
nicolasff, line 292: would it make sense to generate the mask just once per client? we'd still be exercising this code but without calling `rand()` all the time.

Also: it would be great to have it as a command line option, maybe 0 → always disabled, 1 → always enabled, 2 → enabled half the time?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
-	event_base_dispatch(wt->base);
-	event_base_free(wt->base);
-	free(ws_handshake);
-	return NULL;
+    char ws_template[] = "GET /.json HTTP/1.1\r\n"
+                "Host: %s:%d\r\n"
+                "Connection: Upgrade\r\n"
+                "Upgrade: WebSocket\r\n"
+                "Origin: http://%s:%d\r\n"
+                "Sec-WebSocket-Key: webdis-websocket-test-key\r\n"
```
nicolasff, line 343: it would be good to validate the header sent back in response to this key.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+    addr.sin_family = AF_INET;
+    addr.sin_port = htons(wt->hi->port);
+    memset(&(addr.sin_addr), 0, sizeof(addr.sin_addr));
+    addr.sin_addr.s_addr = inet_addr(wt->hi->host);
+
+    ret = connect(fd, (struct sockaddr*)&addr, sizeof(struct sockaddr));
+    if(ret != 0) {
+        fprintf(stderr, "connect: ret=%d: %s\n", ret, strerror(errno));
+        return NULL;
+    }
```
nicolasff, line 366: should we call `ioctl(fd, FIONBIO, &int_one);` after connecting?

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+        workers[i].id = i;
+        workers[i].msg_target = msg_target;
+        workers[i].hi = &hi;
+        workers[i].verbose = verbose;
+        workers[i].state = WS_INITIAL;
+        workers[i].debug = verbose ? debug_verbose : debug_noop;
+        workers[i].timeout_seconds = timeout_seconds;
+        if (thread_count == 1) {
+            printf("Single-threaded mode\n");
+            worker_main(&workers[0]);
```
nicolasff, line 509: I can't tell why this would make a difference vs calling `pthread_create` with a single thread. Debugging… maybe? I'm not even sure this matters.

tests/websocket.c
```diff
@@ -8,319 +8,537 @@
...
+        total += workers[i].msg_received;
+        total_bytes += workers[i].byte_count;
+    }
+
+    /* timing */
+    clock_gettime(CLOCK_MONOTONIC, &t1);
+    float mili0 = t0.tv_sec * 1000 + t0.tv_nsec / 1000000;
+    float mili1 = t1.tv_sec * 1000 + t1.tv_nsec / 1000000;
+
+    if(total != 0) {
```
nicolasff, line 530: minor, but related to the spacing comment at the top of this file: most `if` blocks in the code base don't have space after `if`. If this much code is changed then I think it's fine to have a space, but let's at least keep consistent.

tests/websocket.html
```diff
@@ -3,150 +3,215 @@
...
+
+		<div class="pure-g">
+			<div class="pure-u-1-8"></div>
+			<div class="pure-u-1-3">
+				<h3>JSON</h3>
+				<form class="pure-form">
+					<fieldset>
+						<div class="pure-g">
+							<div class="pure-u-2-3"><label class="ws-state pure-u-23-24" id="json-state">State: Disconnected</label></div>
+							<div class="pure-u-1-3"><button type="submit" class="pure-u-23-24 pure-button pure-button-primary" id="json-btn-connect">Connect</button></div>
```
nicolasff, line 54: what's the point of having 2 levels of nesting here? 2/3 × 23/24… is it for spacing? doesn't PureCSS already include spaces between elements?
In this domain I've only used [Skeleton.css](https://getskeleton.com/), and it did include some spaces between elements.

tests/websocket.html
```diff
@@ -3,150 +3,215 @@
...
+						</div>
+					</fieldset>
+				</form>
+
+				<div class="pure-g">
+					<div class="pure-u-2-3">&nbsp;</div>
+					<div class="pure-u-1-3"><button disabled type="submit" class="pure-u-23-24 pure-button pure-button-primary" id="raw-btn-clear">Clear logs</button></div>
+					<div class="log pure-u-1-1" id="raw-log">
+					</div>
+				</div>
```
nicolasff, line 122: is there a way to re-generate the whole "raw" section without copying & pasting the whole thing twice? :-/

tests/websocket.html
```diff
@@ -3,150 +3,215 @@
...
+	constructor(type, getSerializer, setSerializer) {
+		this.type = type;
+		this.getSerializer = getSerializer;
+		this.setSerializer = setSerializer;
+		this.ws = null;
+
+		$(`${this.type}-btn-connect`).addEventListener('click', event => {
+			event.preventDefault();
+			console.log('Connecting...');
+			this.ws = new WebSocket(`ws://${ host }:${ port }/.${ this.type }`);
```
nicolasff, line 144: I assume that `MozWebSocket` is no longer a thing?

webdis.json
```diff
@@ -26,6 +26,6 @@
 		}
 	],
 
-	"verbosity": 6,
+	"verbosity": 4,
```
nicolasff, line 29: hmm… what if people upgrade and keep their previous `webdis.json`? Then they'll start logging a ton of stuff at TRACE.

src/slog.c
```diff
@@ -76,13 +76,23 @@ slog_fsync_init(struct server *s) {
...
 /**
  * Write log message to disk, or stderr.
  */
 void
 slog(struct server *s, log_level level,
 		const char *body, size_t sz) {
 
+	if(level > s->cfg->verbosity) return; /* too verbose */
+
 	const char *c = "EWNID";
```
nicolasff, line 96: doesn't this need a `T` at the end?

nicolasff (changes requested):
This looks good! I did have quite a few questions though, especially around all the logging. With only a _single_ actual code change that I could find, it doesn't make a huge amount of sense to me to include all this tracing code; it's not really that it will have a performance impact but I'm more concerned about the increase volume of code, many of those have a massive block just to log one line :-/

**edit:** the web page looks awesome!

tests/websocket.c
jessie-murray: part of it, yes.

tests/websocket.c
jessie-murray: It won't happen with the current messages we send, but if we start sending more than one at a time then we'll have the code for it. If we remove this it will make this part of the code dependent on the Redis commands, in a non-obvious way.

tests/websocket.c
jessie-murray: The mask [should not be reused](https://datatracker.ietf.org/doc/html/rfc6455#page-33):
>  The masking key is a 32-bit value chosen at random by the client.
>  When preparing a masked frame, the client MUST pick a fresh masking
>   key from the set of allowed 32-bit values.

tests/websocket.html
jessie-murray: Firefox 90:
image

src/websocket.c
jessie-murray: Yes, this is the actual fix.

tests/websocket.html
jessie-murray: Without this 1/24th grid spacing, the elements all touch. I don't think the extra tags are a major problem in this file that had not been touched for years, it's not frequently edited.

tests/websocket.c
jessie-murray: It doesn't happen since this callback is registered with a single type of event. I have removed this block and the one for reads.

tests/websocket.c
jessie-murray: I've added this validation in commit 3984fb9.

jessie-murray:
@nicolasff the new pub-sub test fails all the time in GitHub actions, I don't think this is due to the WebSocket changes.

webdis.json
jessie-murray: I changed it to 8, this is higher than in previous JSON files.

tests/websocket.html
jessie-murray: Yes, I wasn't too sure about generating the page since it would require doing it in JavaScript, but this is a JS demo anyway. I made the change in 4c0d3f4

jessie-murray:
There is one major issue I found with the server-side WS code: the handshake is sent back to the client in `ws_handshake_reply` with just a single `write()` call: https://github.com/nicolasff/webdis/blob/master/src/websocket.c#L162:

```c
	/* send data to client */
	ret = write(c->fd, buffer, sz);
	(void)ret;
	free(buffer);

	return 0;
}
```

This needs to be added to a buffer and handled in as many chunks as it takes by the event loop.

jessie-murray:
Regarding [this earlier comment](https://github.com/nicolasff/webdis/pull/198#issuecomment-882131930), I made a change to fix it in 8f70db1 and verified it using Valgrind (no memory is leaked or accessed improperly).

jessie-murray:
@nicolasff Thanks for the first review! I think I've either replied to or made changes to address all your comments, and I no longer have any changes left on my TODO list. As for the long list of commits, I know it's a lot so let me know if you want me to rebase them.

src/slog.c
```diff
@@ -121,3 +128,14 @@ slog(struct server *s, log_level level,
 
 	(void)ret;
 }
+
+/**
+ * This wrapper around slog_internal that first checks the log level.
```
nicolasff, line 133: minor: typo in the comment, either an extra "This" or an extra "that"? :-)

src/worker.c
```diff
@@ -35,7 +35,6 @@ worker_new(struct server *s) {
 
 void
 worker_can_read(int fd, short event, void *p) {
-
```
nicolasff, line 38: really minor, but there's no need to change this file.

tests/Makefile
```diff
@@ -1,11 +1,26 @@
...
+	DEBUG_FLAGS += -O0
+	ifeq ($(shell cc -v 2>&1 | grep -cw 'gcc version'),1) # GCC used: add GDB debugging symbols
+		DEBUG_FLAGS += -ggdb3
+	else ifeq ($(shell gcc -v 2>&1 | grep -cw 'clang version'),1) # Clang used: add LLDB debugging symbols
+		DEBUG_FLAGS += -g3 -glldb
+	endif
+else
+	DEBUG_FLAGS += -O3
+endif
+CFLAGS += $(DEBUG_FLAGS)
```
nicolasff, line 18: I'll see if there's a way to put this in common with the other Makefile, it doesn't have to be in this PR though.

tests/websocket.c
```diff
@@ -28,147 +48,412 @@ struct worker_thread {
...
-	if(wt->msg_received % 10000 == 0) {
-		printf("thread %u: %8d messages left (got %9d bytes so far).\n",
-			(unsigned int)wt->thread,
-			wt->msg_target - wt->msg_received, wt->byte_count);
+void
+process_message(struct worker_thread *wt, size_t sz) {
+	if (wt->msg_received && wt->msg_received % 1000 == 0) {
+		printf("thread %d: %8d messages left (got %9d bytes so far).\n",
+		       wt->id,
+		       wt->msg_target - wt->msg_received, wt->byte_count);
```
nicolasff, line 150: sorry I didn't think about it earlier, but this might make more sense as a sleeping thread that reports at a fixed frequency.

tests/websocket.c
```diff
@@ -28,147 +48,412 @@ struct worker_thread {
...
+		wt->cur_hdr_key_len += length;
+	} else { /* first call for this header name */
+		free(wt->cur_hdr_key); /* free the previous header name if there was one */
+		wt->cur_hdr_key_len = length;
+		wt->cur_hdr_key = calloc(length + 1, 1);
+		memcpy(wt->cur_hdr_key, at, length);
+	}
+	wt->debug("%s appended header name data: currently [%.*s]\n", __func__,
+		  (int)wt->cur_hdr_key_len, wt->cur_hdr_key);
+	// wt->cur_header_is_ws_resp = (strncasecmp(at, "Sec-WebSocket-Accept", 20) == 0) ? 1 : 0;
```
nicolasff, line 316: this looks left over from a previous commit.

tests/websocket.c
```diff
@@ -28,147 +48,412 @@ struct worker_thread {
...
+
+	wait_for_possible_write(wt);
+}
+
+static int
+ws_on_message_complete(http_parser *p) {
+	struct worker_thread *wt = p->data;
+
+	wt->debug("%s (wt=%p), upgrade=%d\n", __func__, wt, p->upgrade);
+	// we've received the full HTTP response now, so we're ready to send frames
```
nicolasff, line 422: minor, but the style of comments is a bit inconsistent in this file. (`//` vs `/* .. */`)

nicolasff (changes requested):
thanks for all the updates! this is looking great. I just had a few small comments left but this is close to mergeable. As for the number of commits… I think would probably get in with a merge commit in case the whole PR needs to be reverted; I'll have to think about it.

tests/websocket.c
jessie-murray: That's added! It took me a while due to the deprecation of POSIX semaphores on macOS but it works on both Mac and Linux now.

jessie-murray:
@nicolasff All changes are in, I'm going to stop now for the day but I think the PR looks better now. I believe the WebSocket test is now good enough to add to the "build-and-test" GitHub Action now, although maybe it could go in once the feature is enabled by default.

nicolasff (approved):
Awesome work! This looks great.

the progress thread is a bit more verbose than I'd like, but it's just test code and I don't really have an alternative to suggest.

I'm merging this for now, I think the next step would be to have pub/sub tests for WS since I remember this being a scenario that had issues in the past. Looking up WS RFCs and implementations this weekend I ran into https://github.com/aaugustin/websockets, and thought it might be a good way to write these tests with much less code than in C (I haven't tried it and don't have an opinion about its quality, but it seems to be kind of the default choice for WS in Python).

PR #199, opened 2021-07-24
WebSocket improvements

jessie-murray:
As discussed earlier, here are some tests for the WebSocket feature and several changes that I had to make to get them to pass.

I started with the [websockets](https://websockets.readthedocs.io/en/stable/) library as suggested, but it was highly dependent on `asyncio` and had some expectations that I didn't think followed the WS RFCs. For example, it expects the `Sec-WebSocket-Accept` header to come immediately after `Upgrade` and `Connection`, even though I could not find any requirement saying this had to be the case. I still made this change in case some Webdis clients use this library, but for the tests themselves I found it easier to use [websocket-client](https://github.com/websocket-client/websocket-client) instead.

The changes:
1. Change order of headers
2. Make `Origin` and `Sec-WebSocket-Origin` optional, as this is [not required for non-browser clients](https://datatracker.ietf.org/doc/html/rfc6455#section-4.1).
3. I noticed that calling `ws.close()` in the test would take about 3 seconds to return, so called `close()` on the server side when we detect it.
4. These disconnections were logged at a high level so I changed that too.
5. There were a few bugs in the reporting thread added at the last minute to `tests/websocket.c`, these are fixed now.
6. I added a new WS-only test that uses JSON and "raw" commands.

I started adding those to `build.yml` for the GitHub Action, but I didn't know if it would be ok since the feature is disabled by default so I left them out for now.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -287,14 +287,14 @@ http_client_read(struct http_client *c) {
 		/* broken link, free buffer and client object */
 
 		/* disconnect pub/sub or WS client if there is one. */
-		if(c->self_cmd && c->self_cmd->ac) {
-			struct cmd *cmd = c->self_cmd;
+		if(c->reused_cmd && c->reused_cmd->ac) {
+			struct cmd *cmd = c->reused_cmd;
 
 			/* disconnect from all channels */
-			redisAsyncDisconnect(c->self_cmd->ac);
-			// c->self_cmd might be already cleared by an event handler in redisAsyncDisconnect
+			redisAsyncDisconnect(c->reused_cmd->ac);
+			// c->reused_cmd might be already cleared by an event handler in redisAsyncDisconnect
 			cmd->ac = NULL;
-			c->self_cmd = NULL;
+			c->reused_cmd = NULL;
 
 			/* delete command object */
 			cmd_free(cmd);
--- a/src/client.c
+++ b/src/client.c
@@ -286,15 +286,15 @@ http_client_read(struct http_client *c) {
 	if(ret <= 0) {
 		/* broken link, free buffer and client object */
 
-		/* disconnect pub/sub client if there is one. */
-		if(c->pub_sub && c->pub_sub->ac) {
-			struct cmd *cmd = c->pub_sub;
+		/* disconnect pub/sub or WS client if there is one. */
+		if(c->self_cmd && c->self_cmd->ac) {
+			struct cmd *cmd = c->self_cmd;
 
 			/* disconnect from all channels */
-			redisAsyncDisconnect(c->pub_sub->ac);
-			// c->pub_sub might be already cleared by an event handler in redisAsyncDisconnect
+			redisAsyncDisconnect(c->self_cmd->ac);
+			// c->self_cmd might be already cleared by an event handler in redisAsyncDisconnect
 			cmd->ac = NULL;
-			c->pub_sub = NULL;
+			c->self_cmd = NULL;
 
 			/* delete command object */
 			cmd_free(cmd);
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -375,13 +375,31 @@ cmd_select_format(struct http_client *client, struct cmd *cmd,
 int
 cmd_is_subscribe(struct cmd *cmd) {
 
-	if(cmd->pub_sub_client) { /* persistent command */
+	if(cmd->pub_sub_client || /* persistent command */
+		cmd_is_subscribe_args(cmd)) { /* checked with args */
 		return 1;
 	}
-	if(cmd->count >= 1 && cmd->argv[0] &&
-		(strncasecmp(cmd->argv[0], "SUBSCRIBE", cmd->argv_len[0]) == 0 ||
-		strncasecmp(cmd->argv[0], "PSUBSCRIBE", cmd->argv_len[0]) == 0)) {
-		return 1;
-	}
+	return 0;
+}
+
+int
+cmd_is_subscribe_args(struct cmd *cmd) {
+
+	if(cmd->count >= 2 &&
+		((cmd->argv_len[0] == 9 && strncasecmp(cmd->argv[0], "subscribe", 9) == 0) ||
+		(cmd->argv_len[0] == 10 && strncasecmp(cmd->argv[0], "psubscribe", 10) == 0))) {
+			return 1;
+		}
+	return 0;
+}
+
+int
+cmd_is_unsubscribe_args(struct cmd *cmd) {
+
+	if(cmd->count >= 2 &&
+		((cmd->argv_len[0] == 11 && strncasecmp(cmd->argv[0], "unsubscribe", 11) == 0) ||
+		(cmd->argv_len[0] == 12 && strncasecmp(cmd->argv[0], "punsubscribe", 12) == 0))) {
+			return 1;
+		}
 	return 0;
 }
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -229,7 +229,7 @@ cmd_run(struct worker *w, struct http_client *client,
 		cmd->ac = (redisAsyncContext*)pool_connect(w->pool, cmd->database, 0);
 
 		/* register with the client, used upon disconnection */
-		client->self_cmd = cmd;
+		client->reused_cmd = cmd;
 		cmd->pub_sub_client = client;
 	} else if(cmd->database != w->s->cfg->database) {
 		/* create a new connection to Redis for custom DBs */
@@ -281,7 +281,7 @@ cmd_run(struct worker *w, struct http_client *client,
 	}
 	/* failed to find a suitable connection to Redis. */
 	cmd_free(cmd);
-	client->self_cmd = NULL;
+	client->reused_cmd = NULL;
 	return CMD_REDIS_UNAVAIL;
 }
 
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -35,11 +35,22 @@ cmd_new(struct http_client *client, int count) {
 	return c;
 }
 
+void
+cmd_free_argv(struct cmd *c) {
+
+	int i;
+	fprintf(stderr, "%s: %p\n", __func__, c);
+	for(i = 0; i < c->count; ++i) {
+		free((char*)c->argv[i]);
+	}
+
+	free(c->argv);
+	free(c->argv_len);
+}
 
 void
 cmd_free(struct cmd *c) {
 
-	int i;
 	if(!c) return;
 
 	free(c->jsonp);
@@ -53,12 +64,7 @@ cmd_free(struct cmd *c) {
 		pool_free_context(c->ac);
 	}
 
-	for(i = 0; i < c->count; ++i) {
-		free((char*)c->argv[i]);
-	}
-
-	free(c->argv);
-	free(c->argv_len);
+	cmd_free_argv(c);
 
 	free(c);
 }
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -22,11 +22,12 @@
 #include <ctype.h>
 
 struct cmd *
-cmd_new(int count) {
+cmd_new(struct http_client *client, int count) {
 
 	struct cmd *c = calloc(1, sizeof(struct cmd));
 
 	c->count = count;
+	c->http_client = client;
 
 	c->argv = calloc(count, sizeof(char*));
 	c->argv_len = calloc(count, sizeof(size_t));
@@ -164,7 +165,7 @@ cmd_run(struct worker *w, struct http_client *client,
 		return CMD_PARAM_ERROR;
 	}
 
-	cmd = cmd_new(param_count);
+	cmd = cmd_new(client, param_count);
 	cmd->fd = client->fd;
 	cmd->database = w->s->cfg->database;
 
@@ -224,7 +225,7 @@ cmd_run(struct worker *w, struct http_client *client,
 		cmd->ac = (redisAsyncContext*)pool_connect(w->pool, cmd->database, 0);
 
 		/* register with the client, used upon disconnection */
-		client->pub_sub = cmd;
+		client->self_cmd = cmd;
 		cmd->pub_sub_client = client;
 	} else if(cmd->database != w->s->cfg->database) {
 		/* create a new connection to Redis for custom DBs */
@@ -276,7 +277,7 @@ cmd_run(struct worker *w, struct http_client *client,
 	}
 	/* failed to find a suitable connection to Redis. */
 	cmd_free(cmd);
-	client->pub_sub = NULL;
+	client->self_cmd = NULL;
 	return CMD_REDIS_UNAVAIL;
 }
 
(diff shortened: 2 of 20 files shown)


jessie-murray:
I added a pub-sub test as discussed in the previous pull request, and unfortunately it fails often due to the order of received messages not matching the order in which they were sent. There is one WS subscriber listening to 2 channels and one WS publisher writing 8 messages to each channel, and the messages are often received out of order.

 I left the test disabled by default but it can be turned on with `PUBSUB=1`:
```
PUBSUB=1 ./ws-tests.py
..received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-0']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-0']}
received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-1']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-1']}
received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-2']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-2']}
received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-3']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-3']}
received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-4']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-5']}
received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-5']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-4']}
received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-7']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-6']}
received: {'SUBSCRIBE': ['message', '94e58cf3-e45c-4c9f-8478-3aeb818532cc', 'message-6']}
received: {'SUBSCRIBE': ['message', '40f5db94-72ea-445a-9823-3eeb6f629a24', 'message-7']}
F..
======================================================================
FAIL: test_publish_subscribe (__main__.TestPubSub)
  File "./ws-tests.py", line 161, in test_publish_subscribe
    self.assertEqual(received_per_channel[channel][i], expected,
AssertionError: 'message-5' != 'message-4'
- message-5
?         ^
+ message-4
?         ^
 : In 40f5db94-72ea-445a-9823-3eeb6f629a24: expected at offset 4 was "message-4", actual was: "message-5"

Ran 5 tests in 0.023s

FAILED (failures=1)
```

The test also passes often, I don't know exactly how often it fails but it's a lot, maybe 20-50% of the time in my attempts. I've always seen it propagate all messages tho, so at least they are not dropped.

Both WebSockets and pub-sub have custom code that's different than regular command processing in Webdis, so tracking this down could take some time. I haven't looked at the pub-sub code yet.

nicolasff:
hmm yes this test makes the issue pretty obvious. I don't remember enough about this code to have even a hunch of what might be happening, although I suspect this has to do with the way events are queued separately using libevent (one for message 4, then one for message 5, but then libevent processes the second one first). If that's the case it could "just" be solved by keeping a list of frames to send and appending to it. I find it surprising that messages would be _re-ordered_ and not just dropped by an overwrite (e.g. `message-4` comes from redis, is wrapped into a frame and added to the `http_client`, then `message-5` overwrites it). This would be a more likely bug, but clearly this isn't the case here.

This shows that the WS feature is not quite ready to be enabled by default yet.

jessie-murray:
The main problem with WebSockets in Webdis is the use of events in `client.c` (using `worker_monitor_input` etc) which makes it difficult to separate WebSocket handling from regular HTTP. These new commits introduce a new `ws_client` object which is attached to the `http_client`, transferring all the processing of client events to `websocket.c`. One major issue I ran into was the duplicate registering of events in libevents, which can make them trigger in an infinite loop which uses a lot of CPU. Now `ws_client` uses `event_base_once` to register events and two fields to keep track of whether an event is already registered or not (`scheduled_read` and `scheduled_write`).
Each `ws_client` also has its own persistent `redisAsyncContext` pointer instead of blocking a pool connection.
During testing I also noticed the Python `websocket-client` library sending "connection close" frames so I added support for those, and also for "ping" frames.

For testing I ran `ws-tests.py` with `PUBSUB=1` and used we `websocket` client in tests. Valgrind does not report any leaks.

There are still at least two more changes left to add before I would consider this stable: the "new" Redis connection needs to have called back before the WS messages can be processed, and the HTML demo should have a pub/sub section.

jessie-murray:
Last commits for today: Webdis is now logging commands coming from WebSocket clients, a bug in `pool_on_disconnect` is fixed, Redis errors are sent to WebSocket clients, and the HTML demo now has a pub/sub example. I think that's all for this pull request, which is already too long. @nicolasff let me know if you feel like doing a walkthrough on Tuple again.

src/websocket.c
```diff
@@ -142,57 +196,76 @@ ws_handshake_reply(struct http_client *c) {
...
+ws_log_cmd(struct ws_client *ws, struct cmd *cmd) {
+	char log_msg[SLOG_MSG_MAX_LEN];
+	char *p = log_msg, *eom = log_msg + sizeof(log_msg) - 1;
+	if(!slog_enabled(ws->http_client->s, WEBDIS_DEBUG)) {
+		return;
+	}
+
+	memset(log_msg, 0, sizeof(log_msg));
+	memcpy(p, "WS: ", 4); /* WS prefix */
+	p += 4;
```
nicolasff, line 251: since this is making assumptions about `SLOG_MSG_MAX_LEN` there could be a compile-time check in `slog.h` that it's at least 4 bytes + enough for the date/time format, etc.

src/websocket.c
```diff
@@ -208,30 +281,41 @@ ws_execute(struct http_client *c, const char *frame, size_t frame_len) {
...
+
+				cmd = ws->cmd; /* replace pointer since we're about to pass it to cmd_send */
 			} else {
-				/* get Redis connection from pool */
-				cmd->ac = (redisAsyncContext*)pool_get_context(c->w->pool);
+				/* copy client info into cmd. */
+				cmd_setup(cmd, c);
+
+				/* First WS command; use Redis context from WS client. */
+				cmd->ac = ws->ac;
```
nicolasff, line 312: here is where `cmd->ac` is set, so I guess it won't leak in `cmd_run`, but it also wouldn't use the correct one. And then one of the two would leak?

src/websocket.c
```diff
@@ -264,26 +353,38 @@ ws_msg_add(struct ws_msg *m, const char *p, size_t psz, const unsigned char *mas
...
 	if(sz < 8) {
-		return WS_READING;
+		return WS_READING; /* need more data */
 	}
+	/* copy into "frame" to process it */
+	frame = malloc(sz);
+	if(!frame) {
+		return WS_ERROR;
+	}
+	evbuffer_remove(ws->rbuf, frame, sz);
```
nicolasff, line 384: the return value of `evbuffer_remove` should be checked.

src/websocket.c
```diff
@@ -302,63 +403,88 @@ ws_parse_data(const char *frame, size_t sz, struct ws_msg **msg) {
 		p = frame + 10 + (has_mask ? 4 : 0);
 		if(has_mask) memcpy(&mask, frame + 10, sizeof(mask));
 	} else {
+		free(frame);
 		return WS_ERROR;
 	}
 
 	/* we now have the (possibly masked) data starting in p, and its length.  */
 	if(len > sz - (p - frame)) { /* not enough data */
+		evbuffer_prepend(ws->rbuf, frame, sz); /* put the whole frame back */
```
nicolasff, line 412: here too, and the other two below.

src/websocket.c
```diff
@@ -302,63 +403,88 @@ ws_parse_data(const char *frame, size_t sz, struct ws_msg **msg) {
...
 		return WS_READING;
 	}
 
-	if(!*msg)
-		*msg = ws_msg_new();
-	ws_msg_add(*msg, p, len, has_mask ? mask : NULL);
-	(*msg)->total_sz += len + (p - frame);
+	if(out_msg) { /* we're extracting the message */
+		struct ws_msg *msg = ws_msg_new(frame_type);
+		ws_msg_add(msg, p, len, has_mask ? mask : NULL);
```
nicolasff, line 419: `ws_msg_add` should also have a return value, which should be checked here.

src/websocket.c
```diff
@@ -302,63 +403,88 @@ ws_parse_data(const char *frame, size_t sz, struct ws_msg **msg) {
...
+		} else if(msg->type == WS_PING) { /* respond to ping */
+			ws_frame_and_send_response(ws, WS_PONG, msg->payload, msg->payload_sz);
+		} else if(msg->type == WS_CONNECTION_CLOSE) { /* respond to close frame */
+			ws->close_after_events = 1;
+			ws_frame_and_send_response(ws, WS_CONNECTION_CLOSE, msg->payload, msg->payload_sz);
+		} else {
+			char format[] =  "Received unexpected WS frame type: 0x%x";
+			char error[(sizeof format)];
+			snprintf(error, sizeof(error), format, msg->type);
+			slog(ws->http_client->s, WEBDIS_WARNING, error, 0);
```
nicolasff, line 466: minor: the call to `slog` could use the return value of `snprintf`.

nicolasff, line 466: it would be good to have a test for this, although it would likely require it to be socket-based; I assume most websocket client libraries don't let users send invalid frames.

src/websocket.c
```diff
@@ -387,21 +513,92 @@ ws_reply(struct cmd *cmd, const char *p, size_t sz) {
...
 
-	/* mark as keep alive, otherwise we'll close the connection after the first reply */
-	r->keep_alive = 1;
-
-	r->out = frame;
-	r->out_sz = frame_sz;
-	r->sent = 0;
-	http_schedule_write(cmd->fd, r);
+	/* send WS frame */
+	ws_schedule_write(ws);
```
nicolasff, line 525: here too the `ws_schedule_write` should return a value.

src/websocket.c
```diff
@@ -387,21 +513,92 @@ ws_reply(struct cmd *cmd, const char *p, size_t sz) {
...
 
 	return 0;
 }
+
+static void
+ws_close_if_able(struct ws_client *ws) {
```
nicolasff, line 531: very minor, formatting: most if not all other functions start with a new line.  Sorry this feels like nitpicking and I don't really care either way as long as it's consistent.

src/websocket.c
```diff
@@ -387,21 +513,92 @@ ws_reply(struct cmd *cmd, const char *p, size_t sz) {
...
+	(void)event;
+
+	ws->scheduled_write = 0;
+
+	/* send pending data */
+	ret = evbuffer_write_atmost(ws->wbuf, fd, 4096);
+
+	if(ret <= 0) {
+		ws_client_free(ws); /* will close the socket */
+	} else if(ret > 0) {
```
nicolasff, line 574: CodeQL flagged this as always true.

nicolasff (changes requested):
This looks good! I like the simplified handling in `websocket.c`. I have a few questions about the changes, most are minor but a couple are a bit concerning especially the potential for leaks of a Redis connection. I can see Valgrind not reporting it if some libevent data structures still point to it.

src/cmd.c
```diff
@@ -224,7 +229,7 @@ cmd_run(struct worker *w, struct http_client *client,
 		cmd->ac = (redisAsyncContext*)pool_connect(w->pool, cmd->database, 0);
```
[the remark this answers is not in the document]
jessie-murray: No, since `websocket.c` calls `cmd_send`, not `cmd_run` which uses the HTTP request.

src/websocket.c
jessie-murray: See response above.

src/slog.c
```diff
@@ -95,8 +95,8 @@ slog_internal(struct server *s, log_level level,
 	time_t now;
 	struct tm now_tm, *lt_ret;
 	char time_buf[64];
-	char msg[124];
-	char line[256]; /* bounds are checked. */
+	char msg[1 + SLOG_MSG_MAX_LEN];
+	char line[2 * SLOG_MSG_MAX_LEN]; /* bounds are checked. */
```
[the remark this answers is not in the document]
jessie-murray: No, it already had over 100 bytes of margin.

src/client.c
```diff
@@ -286,15 +286,15 @@ http_client_read(struct http_client *c) {
 	if(ret <= 0) {
 		/* broken link, free buffer and client object */
 
-		/* disconnect pub/sub client if there is one. */
-		if(c->pub_sub && c->pub_sub->ac) {
-			struct cmd *cmd = c->pub_sub;
+		/* disconnect pub/sub or WS client if there is one. */
+		if(c->self_cmd && c->self_cmd->ac) {
```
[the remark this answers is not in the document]
jessie-murray: This can still happen with HTTP-based pub/sub.

jessie-murray:
I changed the calls to `ntohll` and `htonll` in the last commit, these are not portable. I've replaced them with two calls to `ntohl`/`htonl` which is more readable than the previous macros and should work everywhere.

### Test Plan:

* `basic.py` ✅ 
* `ws-tests.py` ✅ 
* `./websocket -c 100 -n 10000` ✅  (10.96 sec, 91,268 msg/sec)
* `./curl-tests.sh` ✅
* `./bench.sh` ✅ (all commands except the long `LRANGE` ran at rates above 91,000 requests/sec)
* `./limits.py` ✅ (no longer reporting a `ResourceWarning`)

All of these succeeded with a simple `make clean all && ./webdis`.

I also ran them under Valgrind, and although they all pass, there was still a part of the Valgrind output that I found a bit concerning (the "still reachable" section):

```
==10== LEAK SUMMARY:
==10==    definitely lost: 0 bytes in 0 blocks
==10==    indirectly lost: 0 bytes in 0 blocks
==10==      possibly lost: 1,232 bytes in 20 blocks
==10==    still reachable: 19,174 bytes in 115 blocks
==10==         suppressed: 0 bytes in 0 blocks
```

It looks like the memory reported as still reachable is due to the Redis connections created by `redisAsyncConnect` for each new WebSocket client. I first tracked all calls to `ws_client_new` and `ws_client_free` and these were called in the expected order. I also added some logging at the main allocation/free points in hiredis and saw all of the memory being freed, so my best guess is that these are due to some internal data structures of libevent that may have grown at runtime but are not actually still usable or still pointing to memory that has clearly been freed.

As I was exploring the hiredis code I noticed that the version embedded in Webdis is pretty old, and the recent hiredis code looks significantly different. It might be worth looking into upgrading at some point.

nicolasff:
Thanks for the details! I ran through a similar test plan and more, also testing it on ARM with no issues (ARM Cortex-A57, little endian like all the machines I have access too but often different enough to surface portability issues).

I had actually tried upgrading hiredis a few months back and gave up due to incompatibility issues I never managed to figure out. I might give it a shot again soon.

I've been through these changes several times now and no longer have any suggestions (thanks for the updates!). With websockets still disabled by default I think this is safe to merge. I'll try coming up with a checklist of what remains before this feature can be turned on by default, I think that means at least:

1. Upgrading hiredis
2. Consolidating the tests to run all of them with a single command
3. Running the tests while webdis runs under valgrind
4. Adding docs

And probably more.

Thanks again for this PR, this looks great.

Issue #201, opened 2021-09-09
REDIS 6 now offers TLS, will Webdis?


cilex-ft:
We manage hundreds of servers, each with their own redis / webdis server. 

Webdis now supports user/password to connect to Redis, so we can consider moving to 1 single Redis server or cluster while ensuring good data isolation between our servers and clients.

However the data flow on the internal network between webdis / redis might be read by an attacker.

We have measures to secure our internal network, such as: we encrypt all the internal traffic (to mysql, between reverse proxies and web servers behind...); all servers run a firewall to avoid attacks from peers; webdis servers only answer on 127.0.0.1; and undisclosed more.

I couldn't find an include to hiredis_ssl.h in webdis code, so it's my understanding that you don't yet support [hiredis ssl/tls](https://github.com/redis/hiredis/tree/v1.0.0#ssltls-support) - am I right?

It would be great if you can consider adding this feature.

nicolasff:
Hi @cilex-ft,

That's correct, webdis doesn't support TLS connections to redis. I see that TLS was only added in Redis 6, so this is relatively new feature that webdis doesn't support.

I completely get how important it is, and it seems like a good idea to add it. I can't really make such changes myself due to IP restrictions linked to my current job, but have been helping a contributor recently who's made a lot of important fixes to webdis so I'll talk to him about what this involves and whether he's interested in taking this on.

Let's keep this open as a reference issue and an eventual PR can link to it.

Thanks for the suggestion!

cilex-ft:
Thanks for delivering this, and sorry for my delay answering, daily operations don't leave much time for experimenting. 
Our choices would be 
1) using a package, or 
2) build if required, then deliver to containers via ansible. 
We don't use Docker (and we won't). 
We should find some time in november to test... sounds great!

nicolasff:
I don't manage the packages for distros, this would be too much work that is better left to folks who are much more familiar with their specific platforms than I could ever be. Some even take care of it very quickly, e.g. 0.1.18 was tagged on [October 22](https://github.com/nicolasff/webdis/releases/tag/0.1.18) and updated in Homebrew just [3 days later](https://github.com/Homebrew/homebrew-core/commit/0062b18bf90d76097a9131b4244cde41f268a3cb).

Either way, thanks @cilex-ft for suggesting this feature and thanks to @jessie-murray for the concise implementation (although a lot of it is from Hiredis).

[...]

PR #202, opened 2021-10-16
SSL support for connections to Redis

jessie-murray:
This PR adds support for encrypted connections to Redis, using Hiredis's SSL support which was added in v1.0.0. This fixes issue #201.
I started by upgrading Hiredis to [v1.0.2](https://github.com/redis/hiredis/releases/tag/v1.0.2) and re-adding the macros that mark some functions as unused. A few fields had also changed types, so I had to update Webdis in 2 places to use the correct `sprintf` format. Once this was integrated I added a block in the Makefile to enable SSL support, which is done with a pre-processor macro. The reason for this is to avoid linking against `libssl` and `libcrypto` by default due to the frequent security vulnerabilities that are found in these libraries. With the macro in place I added SSL initialization code in `server.c` and handshake in `pool.c`, as well as a new config block read from `conf.c`. Lastly I documented the configuration and added a troubleshooting section for some issues.

@nicolasff I see there has been no response to your Hiredis question but after trying the different approaches we talked about and with the sentinel example I'm more confident now that this is the right way to use this feature.

The main question left is how this would be packaged with Docker: if the goal of this build option is to not expose a binary built with SSL by default, then how would someone connect to Redis securely with this image? The issue is not even to have the binary in the image, but also to include all the PEM files. Maybe Webdis can have a Docker image with SSL and one without?

I used [this guide](https://nishanths.svbtle.com/setting-up-redis-with-tls) to setup SSL and ran all the tests including Websockets with SSL enabled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ Follow this table to diagnose issues with SSL connections to Redis.
 * Multi-threaded server, configurable number of worker threads.
 * [WebSocket support](#websockets) (Currently using the specification from [RFC 6455](https://datatracker.ietf.org/doc/html/rfc6455)).
 * Connects to Redis using a TCP or UNIX socket.
-* Support for [secure connections to Redis](https://redis.io/topics/encryption) (requires Redis 6 or newer).
+* Support for [secure connections to Redis](#configuring-webdis-with-ssl) (requires [Redis 6 or newer](https://redis.io/topics/encryption)).
 * Restricted commands by IP range (CIDR subnet + mask) or HTTP Basic Auth, returning 403 errors.
 * Support for Redis authentication in the config file: set `redis_auth` to a single string to use a password value, or to an array of two strings to use username+password auth ([new in Redis 6.0](https://redis.io/commands/auth)).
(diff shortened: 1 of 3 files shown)


nicolasff:
I forgot to mention, I'm not really sure how to provide this feature. I'm kind of weary of having it enabled by default, I feel like there has been so many CVEs on Alpine due to openssl vulnerabilities that I would really prefer not to link against it if it's not needed. I mean just look at the last few releases, almost all of them bump the version of the Alpine base image due to vulnerabilities, many of them in openssl.

Maybe 2 binaries could work? e.g. `/usr/bin/webdis` and `/usr/bin/webdis-ssl`? Then since users who want to use SSL have to build their own image anyway with a custom config file that points to the right certs and keys, they can also overwrite `webdis` with `webdis-ssl` or use it as their entry point.

I'm also not so sure about doing it that way, because maybe it's not a big deal if there's an openssl vulnerability in a binary where it's unused. So yeah, none of this really changes anything about this PR, but I think there's still a bit more to do once it's merged and it's not completely clear to me yet.

If anyone reading this has a strong opinion about it, suggestions would be welcome :-)

PR #205, opened 2021-11-06
Reduce malloc/free usage for HTTP responses

jessie-murray:
This pull request makes two changes: the first to avoid copying header strings for `http_response` when static values are used, and the second to avoid re-allocating the headers array for each header being added to it.

The value of 9 for `DEFAULT_HEADERS_ARRAY_SIZE` was found with a `curl` to `/PING`, which returned 9 headers.

I ran Webdis under valgrind to count the allocations produced by `./basic.py`.

Before:
```
==10== HEAP SUMMARY:
==10==     in use at exit: 22,404 bytes in 233 blocks
==10==   total heap usage: 4,122 allocs, 3,889 frees, 340,545 bytes allocated
```

After:
```
==10== HEAP SUMMARY:
==10==     in use at exit: 22,916 bytes in 233 blocks
==10==   total heap usage: 3,000 allocs, 2,767 frees, 281,176 bytes allocated
```

This reduced the number of `malloc` calls by over 27%, and `free` by the same count. The same number of blocks were left allocated after the test and both executions reported 0 memory leaks.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -11,6 +11,14 @@
 
 /* HTTP Response */
 
+#define DEFAULT_HEADERS_ARRAY_SIZE 9
+
+static void
+http_response_allocate_headers(struct http_response *r) {
+	r->headers_array_size = DEFAULT_HEADERS_ARRAY_SIZE;
+	r->headers = calloc(r->headers_array_size, sizeof(struct http_header));
+}
+
 struct http_response *
 http_response_init(struct worker *w, int code, const char *msg) {
 
@@ -25,6 +33,14 @@ http_response_init(struct worker *w, int code, const char *msg) {
 	r->msg = msg;
 	r->w = w;
 	r->keep_alive = 0; /* default */
+
+	/* pre-allocate array for headers */
+	http_response_allocate_headers(r);
+	if(!r->headers) {
+		if(w && w->s) slog(w->s, WEBDIS_ERROR, "Failed to allocate http_response headers", 0);
+		free(r);
+		return NULL;
+	}
 
 	http_response_set_header(r, "Server", "Webdis", HEADER_COPY_NONE);
 
@@ -57,6 +73,14 @@ http_response_init_with_buffer(struct worker *w, char *data, size_t data_sz, int
 	}
 	r->w = w;
 
+	/* pre-allocate array for headers */
+	http_response_allocate_headers(r);
+	if(!r->headers) {
+		if(w && w->s) slog(w->s, WEBDIS_ERROR, "Failed to allocate http_response headers", 0);
+		free(r);
+		return NULL;
+	}
+
 	/* provide buffer directly */
 	r->out = data;
 	r->out_sz = data_sz;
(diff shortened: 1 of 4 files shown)


src/http.c
```diff
@@ -77,29 +101,42 @@ http_response_set_header(struct http_response *r, const char *k, const char *v)
...
+			if(r->headers[i].copy == HEADER_COPY_KEY) free(r->headers[i].key);
+			if(r->headers[i].copy == HEADER_COPY_VALUE) free(r->headers[i].val);
 			break;
 		}
 	}
 
 	/* extend array */
-	if(pos == r->header_count) {
+	if(pos == r->headers_array_size) {
+		/* FIXME: allocation could fail */
```
nicolasff, line 112: hmm thanks for pointing this out, this was never handled.

src/http.c
```diff
@@ -77,29 +101,42 @@ http_response_set_header(struct http_response *r, const char *k, const char *v)
 		if(strncmp(r->headers[i].key, k, key_sz) == 0) {
 			pos = i;
 			/* free old value before replacing it. */
-			free(r->headers[i].key);
-			free(r->headers[i].val);
+			if(r->headers[i].copy == HEADER_COPY_KEY) free(r->headers[i].key);
+			if(r->headers[i].copy == HEADER_COPY_VALUE) free(r->headers[i].val);
 			break;
 		}
 	}
```
nicolasff, line 108: Looking at the callers of `http_response_set_header` it feels like all these calls to `strncmp` are probably useless for almost all responses, if not literally all. It could make sense to have a way to skip it if we know the header been added has never been set.

src/http.c
```diff
@@ -126,8 +163,8 @@ http_response_cleanup(struct http_response *r, int fd, int success) {
 
 	/* cleanup response object */
 	for(i = 0; i < r->header_count; ++i) {
-		free(r->headers[i].key);
-		free(r->headers[i].val);
+		if(r->headers[i].copy & HEADER_COPY_KEY) free(r->headers[i].key);
+		if(r->headers[i].copy & HEADER_COPY_VALUE) free(r->headers[i].val);
```
nicolasff, line 167: hmm these use `&` while the ones above use `==`. It does work with the current values assigned to these enum entries, but there's no `HEADER_COPY_BOTH` for example that would have both bits set. It would be better if the two uses were consistent.

nicolasff (changes requested):
Sorry for the delay in reviewing this. Impressive numbers! I had just a few small suggestions.

nicolasff:
This looks great, thanks for adding the new flag.

Since the original test was just with `basic.py`, I ran the same comparison with `ws-tests.py`, getting "7,767 allocs, 7,530 frees, 28,872,327 bytes allocated" with the latest release and "7,767 allocs, 7,530 frees, 28,873,671 bytes allocated" with this pull request.

I was surprised to see the same number of allocations, before realizing this made total sense given that `ws_handshake_reply` builds the response with `HTTP/1.1 101 Switching Protocols` and not the code in `http.c`.

With just `basic.py`, I see the the same improvements as those reported above, "4,123 allocs, 3,890 frees, 341,057 bytes allocated" on 0.1.18 vs "3,000 allocs, 2,767 frees, 281,176 bytes allocated" with this PR.

Well done!

nicolasff (approved):

PR #210, opened 2021-12-05
Only call ws_client_free once all scheduled events have triggered

### LINKED ISSUE #209: core dumped when subscrib to redis
Thanks for your help.

When I use websocket to subscribe to redis, as long as I close the browser, the server will report an error and stop.

        tickSocket.onopen = function() {
          console.log("tick socket connected!");
          tickSocket.send(JSON.stringify(["SUBSCRIBE","data_stream"]));
        };

root@VM-0-6-ubuntu:/usr/local/webdis# ./webdis webdis.json
[err] buffer.c:565: Assertion buffer->refcnt > 0 failed in evbuffer_decref_and_unlock_
Aborted (core dumped)

jessie-murray:
Fixes #209. A WS client socket closure could cause Webdis to schedule the send of a closing frame, leading to both `EV_READ` and `EV_WRITE` scheduled events. They would both fail and each lead to a call to `ws_client_free`, causing a double-free and a crash.

### DIFF
--- a/src/websocket.c
+++ b/src/websocket.c
@@ -113,7 +113,7 @@ ws_client_new(struct http_client *http_client) {
 	return ws;
 }
 
-void
+static void
 ws_client_free(struct ws_client *ws) {
 
 	/* mark WS client as closing to skip the Redis callback */
@@ -558,7 +558,7 @@ ws_frame_and_send_response(struct ws_client *ws, enum ws_frame_type frame_type, 
 	return ws_schedule_write(ws);
 }
 
-static void
+void
 ws_close_if_able(struct ws_client *ws) {
 
 	ws->close_after_events = 1; /* note that we're closing */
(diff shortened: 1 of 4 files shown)


nicolasff:
Thanks for the quick PR!

Looking at the other call to `ws_client_free` (from `worker_can_read`), shouldn't it also be changed to `ws_close_if_able`?
Also, this feels like the sort of bug that could easily be covered by a new test in `ws-tests.py`. It seems like it's also important to check for leaks for any change dealing with memory allocation.

And lastly the CodeQL results show 4 new warnings. Although these seem unrelated to this particular change, they are worth investigating.

nicolasff:
Looks good!

The new test does reproduce the issue reliably, and the fix solves it. I also confirmed that this is no longer reproducible with a browser. Valgrind reports no leaks.

Thanks again!

PR #231, opened 2023-02-20
Compile in C99 mode

baonq-me:
Fix error when compiling in GCC 4.8.5 on CentOS 7

```
src/websocket.c: In function ‘ws_log_cmd’:
src/websocket.c:247:2: error: ‘for’ loop initial declarations are only allowed in C99 mode
  for(int i = 0; p < eom && i < cmd->count; i++) {
  ^
src/websocket.c:247:2: note: use option -std=c99 or -std=gnu99 to compile your code
make: *** [src/websocket.o] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ CFLAGS ?= -Wall -Wextra -Isrc -Isrc/jansson/src -Isrc/http-parser -MD
 LDFLAGS ?= -levent -pthread
 
 # Pass preprocessor macros to the compile invocation
-CFLAGS += $(CPPFLAGS)
+CFLAGS += $(CPPFLAGS) -std=gnu99
 
 # check for MessagePack
 MSGPACK_LIB=$(shell ls /usr/lib/libmsgpack.so 2>/dev/null)


nicolasff:
Hi @baonq-me,

Thanks for the pull request! I was able to reproduce the issue with this docker image of CentOS 7 and `gcc.x86_64` version `4.8.5-44.el7`.

That said, I don't think this change is quite right. `-std=gnu99` is not the same as C99, see the description for it in the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Standards.html):

> `-std=gnu99` (for C99 with GNU extensions)

From what I can tell, clang does not fully support `-std=gnu99` and Webdis does not use any of these extensions either. `-std=c99` would make sense though, as you point out this construct requires it.

Thinking about this change, it seems to me that it would make sense to do this where most of the compile flags are already set. The line you changed is there just to bring in `CPPFLAGS` as the comment above explains, so with your change this comment is no longer accurate.

You can put it here:
```diff
- CFLAGS ?= -Wall -Wextra -Isrc -Isrc/jansson/src -Isrc/http-parser -MD
+ CFLAGS ?= -std=c99 -Wall -Wextra -Isrc -Isrc/jansson/src -Isrc/http-parser -MD
```

Can you please make this change? I'll run the validation workflows once it's in.

nicolasff:
Hmm, I see that `-std=c99` is actually not enough to build on Centos 7:

```
src/server.c: In function 'server_daemonize':
src/server.c:203:61: error: 'O_NOFOLLOW' undeclared (first use in this function)
   int pid_fd = open(pidfile, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0600);
```
If you're up for it, would you mind removing `O_NOFOLLOW` from this function call? It was only [standardized in POSIX.1-2008](https://man7.org/linux/man-pages/man2/open.2.html#:~:text=This%20flag%20is%20a%20FreeBSD%20extension%2C%20which%20was%20added%20to%20Linux%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20in%20version%202.1.126%2C%20and%20has%20subsequently%20been%20standardized%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20in%20POSIX.1%2D2008.) which is a bit too recent for my liking.

So please add this change to your pull request, in `src/server.c`:
```diff
-               int pid_fd = open(pidfile, O_WRONLY | O_CREAT | O_TRUNC | O_NOFOLLOW, 0600);
+               int pid_fd = open(pidfile, O_WRONLY | O_CREAT | O_TRUNC, 0600);
```

With the `-std=c99` change shown above and the removal of `O_NOFOLLOW`, I was able to build Webdis on the latest `main` (c6b9d5221365d489a15ce491911f2d10e9aed20e) albeit with some warnings. I'll look into those soon.

nicolasff:
@baonq-me I've pulled in your commit here: 50d16b16ca4724373e02edffdfe08709dae52450 along with a couple more CentOS fixes. You should be able to build Webdis cleanly from `master` on CentOS 7, please let me know if this is not the case.

I've tested this with a Docker image for CentOS 7, Webdis builds and basic tests are passing.

Issue #255, opened 2025-03-13
Content-Encoding: gzip


aminroosta:
Is it possible to SET and GET gzip or brotli encoded JSON?

nicolasff:
Hello,

It _is_ technically possible to set and retrieve gzip-encoded JSON, but this does not mean that Webdis will compress a regular JSON output. So to be clear, it does not mean that you can have Webdis respond with a compressed version of its regular responses.
Instead, the only way to do this is really independent of JSON, and involves **storing binary data** into Redis with Webdis fetching it as it was stored. The advantage is that you use less memory in Redis, but the downside is that Webdis is then used to simply fetch binary data instead of giving you JSON.

There is [an example](https://github.com/nicolasff/webdis#file-upload) in the `README` showing how to store a PNG file in Redis via Webdis. You do get a standard `{"SET":[true,"OK"]}` when you upload the file, but then you'd need to fetch its binary contents rather than get a JSON response like `{"GET":"(some data here)"}`.

Here's a demo showing a gzip'd file stored in Redis via Webdis, similar to the PNG example. Let's say we have a JSON file containing some record (really, it could be any kind of file):
```
$ wc -c data.json 
     408 data.json

$ head -4 data.json
{
    "id": 834719,
    "username": "johndoe",
    "email": "johndoe92@example.com",
...
```
It's 408 bytes raw, so let's gzip it and store it under the key `user:johndoe` in Redis, via Webdis:
```
$ curl -v --upload-file <(cat data.json | gzip -) http://127.0.0.1:7379/SET/user:johndoe
> PUT /SET/user:johndoe HTTP/1.1
...
{"SET":[true,"OK"]}
```

This command uses [process substitution](https://en.wikipedia.org/wiki/Process_substitution), so `<(cat data.json | gzip -)` is a _file_ (a pipe, really) containing the compressed data resulting from the command. You can always do this in two steps by compressing the file first and then sending the compressed file.

The original file was 408 bytes, but let's see how many bytes are now stored at that key:
```
$ curl -s http://127.0.0.1:7379/STRLEN/user:johndoe              
{"STRLEN":256} 
```
So we _are_ storing compressed data. Let's now need to fetch this binary data _alone_, meaning not wrapped in a `{"GET": ...}` response. You can [choose the exact content-type](https://github.com/nicolasff/webdis#custom-content-type) that Webdis will return, but for this demo we'll just add `.bin` at the end of the key to get it as `binary/octet-stream`.

[...]

PR #256, opened 2025-03-16
Adds support for the Content-Encoding header

aminroosta:
Fixes https://github.com/nicolasff/webdis/issues/255

### DIFF
--- a/src/cmd.c
+++ b/src/cmd.c
@@ -321,27 +321,31 @@ cmd_select_format(struct http_client *client, struct cmd *cmd,
 		size_t sz;
 		formatting_fun f;
 		const char *ct;
+		const char *ce;
 	};
 	struct reply_format funs[] = {
-		{.s = "json", .sz = 4, .f = json_reply, .ct = "application/json"},
-		{.s = "raw", .sz = 3, .f = raw_reply, .ct = "binary/octet-stream"},
+		{.s = "json", .sz = 4, .f = json_reply, .ct = "application/json", .ce = NULL},
+		{.s = "raw", .sz = 3, .f = raw_reply, .ct = "binary/octet-stream", .ce = NULL},
 
 #ifdef MSGPACK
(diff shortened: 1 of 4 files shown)


nicolasff:
Hi Amin, thanks for the PR! I've only given it a brief look so far but will spend some time on it today.

In the meantime I did trigger the tests which run on a range of systems, they're all marked as failed simply because the same test failed in all environments; so it looks all red but it is really just about one test :-)

From the output:

```
FAIL: test_encoding (__main__.TestJSON.test_encoding)
success type (+OK)
  File "/__w/webdis/webdis/./tests/basic.py", line 87, in test_encoding
    self.assertTrue(f.getheader('ETag') == '"fcb2917ae[14](https://github.com/nicolasff/webdis/actions/runs/13878036214/job/38846608045?pr=256#step:9:15)a62e911fbfb2b8ea66800"')
AssertionError: False is not true
```

To run this validation yourself, start Webdis with websockets enabled (set `"websockets": true` in `webdis.json`):
```diff
-  "websockets": false,
+  "websockets": true,
```

and in the `tests/` directory run `./basic.py` and `./ws-tests.py`, that's usually how I run an initial check over Webdis.


You'll need to install `websocket-client>=1.1.0` for Python, I suggest you create a venv to leave your global Python install untouched. With Redis and Webdis already running:
```shell
cd tests
python -m venv .venv
source .venv/bin/activate
pip install -r requirements.txt

./basic.py
./ws-tests.py
```

aminroosta:
Hello @nicolasff, I appreciate you taking the time to review the PR. 
I've made a tiny change in https://github.com/nicolasff/webdis/pull/256/commits/ec025f8dee073ac564a4c8672e0197d7ff49144b to support both `key.json.gzip` and `key.gzip`. 
I hope the pointer arithmetic is accurate 🤞.

aminroosta:
Will fix the unit tests later today. The python gzip implementation includes a timestamp.

aminroosta:
Done.

